PR #129, opened 2026-09-02
feat: Expose date1904 property on ReadOnlyWorkbook

### LINKED ISSUE #128: feat: expose date epoch for streaming / support converting date serials during streaming
## Problem

I want to use `.iterRows()` to stream a worksheet. For date values, I want to format them, for example, as RFC3339/ISO8601 strings. Currently, `loadWorkbookStream` returns a `ReadOnlyWorkbook` and `iterRows` returns `ReadOnlyCell`s. The `ReadOnlyWorkbook` provides a `Stylesheet`, so I can determine that the cell is date-formatted. However, without the workbook's epoch, I don't think it's possible to correctly convert that, using e.g. `excelToDate`

## Existing paths considered

As far as I can tell, I'd need to separately re-open the entire workbook just to get the `date1904` property so I can determine the correct date format.

## Proposed solution

Add `date1904` to `ReadOnlyWorkbook` mirroring `Workbook`

```ts
export interface ReadOnlyWorkbook {
  sheetNames: string[];
  styles: Stylesheet;
  /** Date1904 mode toggles between Excel's two epoch systems. */
  date1904: boolean;
```

## Alternatives

1. Expose epoch on `ReadOnlyWorkbook` instead. This is inconsistent with `Workbook`
2. Introduce more involved APIs to stream already-converted/typed values. Easier for me, but still requires `date1904` on the `ReadOnlyWorkbook` at least internally and may violate your "one way to do things" principle.
3. Call `loadWorkbook` to get `date1904`. This defeats the purpose of calling `loadWorkbookStream`

## Scope and compatibility

Additive and backwards compatible.

## Confirmation

- [x] I have read the project's "one way to do one thing" policy and confirmed
      this proposal is not a parallel path to an existing capability.
- [x] I have searched existing issues (open and closed) and confirmed this is
      not a duplicate.
- [x] I have a real, specific use case for this — not a hypothetical "could be
      useful."
- [x] If I used an LLM to draft this issue, I have read and verified every
      claim, including the "existing paths" section, and am willing to defend
      the proposal in follow-up.

matt-felicity:
## Summary

This adds `date1904` property to `ReadOnlyWorkbook` so that streaming readers can select the correct epoch when converting these date-styled serials.

## Motivation

When streaming with `.iterRows()` for example, it's not possible to correctly convert date-styled cells to a useful value without knowing the workbook epoch type, which was not available through these APIs.

Closes #128 

## Changes

- Adds `date1904` property to `ReadOnlyWorkbook`
- Exposes `parseDate1904` to allow `loadWorkbookStream` to parse it

## Testing

- I added new tests to show how this is used with streaming reads

## Breaking changes

None

## Checklist

- [ ] I have added or updated documentation where user-visible behavior changed.
      the result.
      marked accordingly.
      comments.

### DIFF
--- a/.changeset/streaming-date-epoch.md
+++ b/.changeset/streaming-date-epoch.md
@@ -1,0 +1,5 @@
+---
+"@office-kit/xlsx": minor
+---
+
+Expose the workbook `date1904` flag from `loadWorkbookStream` so streamed Excel date serials can be converted with the correct epoch.
--- a/src/io/load.ts
+++ b/src/io/load.ts
@@ -175,10 +175,10 @@ const FUNCTION_GROUP_TAG = `{${SHEET_MAIN_NS}}functionGroup`;
 /**
  * Parse the `<workbookPr date1904>` flag. Mac-origin workbooks set
  * `date1904="true"`; everything else uses the Windows 1900 epoch. The value
- * drives Date / Duration cell serial conversion in worksheet/writer.ts and
- * reader.ts.
+ * drives Date serialization and lets readers select the epoch when converting
+ * numeric date serials.
  */
-function parseDate1904(workbookRoot: XmlNode): boolean {
+export function parseDate1904(workbookRoot: XmlNode): boolean {
   const pr = findChild(workbookRoot, WORKBOOK_PR_TAG);
   if (!pr) return false;
   const v = pr.attrs['date1904'];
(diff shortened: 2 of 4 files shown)


matt-felicity:
I am not sure if there's a place where this makes sense to be documented, but happy to do that if so.

baseballyama (approved):
Thank you~~~!!!

PR #133, opened 2026-09-15
fix: keep types resolvable on moduleResolution node16 / nodenext

kibertoad:
## Summary

Relative imports inside the shipped `.d.ts` files carried no file extension, so consumers on `moduleResolution: node16` or `nodenext` silently lost every type this library exports. The declarations now carry `.js`, the repo's own type-check resolves with `nodenext` so the compiler rejects a missing extension where it is written, and a new CI job checks the packed tarball with `are-the-types-wrong` plus a stand-in downstream project that compiles and runs
against it.

## Motivation

`tsc` copies relative specifiers verbatim into its output, and `src/` used the extensionless form throughout. `dist/io/index.d.ts` shipped `export { loadWorkbook } from './load'`, which Node's ESM rules reject, so all 390 relative specifiers across the declaration tree were unresolvable under `node16` / `nodenext`.

Previously failure was silent, which is why it survived to `0.11.0`. TS2834 is raised inside `node_modules`, where the near-universal `skipLibCheck: true` discards it, and the affected declarations degrade to TypeScript's error type.
Reproduced against `0.11.0` with `module: node16`, `strict: true`, `skipLibCheck: true`:

```ts
import { createWorkbook } from '@office-kit/xlsx/workbook';
const broken: number = createWorkbook(); // exit 0, no error
```

With `skipLibCheck: false` the same project reports 31 TS2834 errors, all inside `node_modules/@office-kit/xlsx/dist`. `attw --profile esm-only` reported `InternalResolutionError` on all 15 subpaths for both `node16 (from ESM)` and `node16 (from CJS)`.

Switching consumers to `moduleResolution: bundler` is not the fix, and the reason shapes the approach here. A project that compiles with plain `tsc` and runs `node dist/server.js` unbundled is still subject to Node's ESM rules at runtime, so `bundler` would let TypeScript accept specifiers Node then rejects, trading a contained typing bug for a class of production `ERR_MODULE_NOT_FOUND`. The fix belongs in the artifact, and the gate has to include a real unbundled `node` run rather than a type-check alone.

## Changes

- All 390 relative specifiers in the shipped `.d.ts` tree now carry `.js`, so
  every subpath resolves under `node16` / `nodenext`. No export name, type
  signature or runtime behavior changed.
- The root `tsconfig.json` resolves with `nodenext` instead of `bundler`, so an
  extensionless relative import in `src/` now fails `pnpm typecheck` with TS2835
  at the line that wrote it.
- Two new scripts, both chained into `prepublishOnly` and a new `package` CI job:
  `pnpm check:attw` runs `are-the-types-wrong` over the packed tarball, and
  `pnpm check:consumer` installs that tarball into `tests/consumer` and compiles
  plus runs it under `node16`, `nodenext` and `bundler` resolution.
- New dev dependency: `@arethetypeswrong/cli`.

## Testing

The existing suite passes unchanged (2319 passing, and the same 36 conformance failures before and after, from `xmllint` being absent locally).

The gate was verified to fail, not just to pass. Three negative controls:

| Injected regression | What went red |
| --- | --- |
| `.js` stripped from all shipped `.d.ts` | `check:attw`, plus all four `node16` / `nodenext` type-check passes. 68 assertion failures from `tests/consumer/src/no-any.ts` under `skipLibCheck: true` |
| `.js` dropped from one specifier in `src/io/index.ts` | `pnpm typecheck`, TS2835 |
| an `exports` entry pointed at a missing file | only the runtime passes, with `ERR_MODULE_NOT_FOUND`. `check:attw` stayed green, which is why the unbundled `node` run is a separate gate |

Detecting the silent case is the part worth review attention. A conditional type over `any`, or over the error type TypeScript substitutes for an unresolved import, takes both branches and collapses to `boolean`, so `Expect<T extends true>` rejects it. Two probes I tried first do not work and are called out in `tests/consumer/src/no-any.ts` so nobody reintroduces them: `keyof`-based checks pass trivially because `keyof any` is `string | number | symbol`, and `0 extends 1 & T` distinguishes `any` but not the error type. The assertions are therefore structural.

`tests/consumer/package.json` carries a different package name on purpose.
Without that boundary, Node and TypeScript both resolve `@office-kit/xlsx/io` through package self-reference, straight back to the repo's own `dist/`, and the fixture would never touch the installed tarball.

Deliberately left out:

- The `.d.ts` tree stays unbundled (`tsdown` still has `dts: false`). Bundled
  declarations would remove internal specifiers entirely, but that is a larger
  change to the shipped type layout and to typedoc's input, and it is not needed
  once the extensions are right.
- `node10` support, which would mean a `typesVersions` map or a CJS build.
- The `package` job runs on ubuntu only. The artifact is platform-independent;
  the cross-platform paths in `scripts/check-consumer.mjs` were exercised
  locally on Windows.
- `publint`, which overlaps with `attw` here and was not part of the ask.

## Breaking changes

None to the API. Worth knowing for the release note: a `node16` / `nodenext` consumer whose build currently passes only because everything from this library is typed as an error type will start seeing real type errors in their own code after upgrading. Those errors were always there, just unreported.

## Checklist

      the result.
      marked accordingly. (Not a breaking change; changeset is `patch`.)
      comments.


baseballyama:
@kibertoad Thank you!!!

PR #135, opened 2026-09-16
feat!: make generated reports cheaper to write and their bytes reproducible

kibertoad:
## Summary

Building a styled, formula-carrying workbook from scratch takes more calls than it
should, and two of those calls destroy work without saying so. This makes the
generate-a-report path cheap, closes both footguns, and pins the one remaining source
of non-determinism in the output so golden-file tests become possible.

## Motivation

No issue; scoped from an analysis of a real five-sheet customer-facing report renderer
built on `0.11.1` (a `payload -> Uint8Array` function, never read back in production).
Everything it needed was reachable, so this is not about missing capability. It is
about how many calls it took and about two behaviours that cost the author real
debugging time:

- **`setCell` blanks the cell.** `setCell(ws, r, c)` reads like "reach the cell at
  (r, c)" and writes `null`. The renderer lost 34 of its 37 formulas that way, because
  a border pass ran over already-populated rows. The library already had the
  get-or-create pattern the renderer had to hand-roll (`applyToRange`, `setRangeStyle`
  and `setRangeBorderBox` all do `ws.rows.get(r)?.get(c) ?? setCell(ws, r, c)`), but it
  was not public. Worse, the cheatsheet and the formula recipe both taught the no-value
  form as *the* way to reach a cell.
- **Replacing a font drops the workbook default.** `setCellFont(wb, c, makeFont({ bold:
  true }))` is a legal call that registers `<font><b/></font>` with no `<sz>` and no
  `<name>`. Excel substitutes its own default, LibreOffice and Sheets may render
  differently, and nothing warns. The renderer wrapped every font change in a local
  `patchFont` helper to work around it.

Beyond that, the renderer's per-row styling loops made roughly 200 style calls for a
16-row sheet, and its determinism test had to compare parsed cells rather than bytes
because every ZIP entry carries a wall-clock timestamp.

## Changes

**Breaking**

- `setCell(ws, row, col, value)` and `setCellByCoord(ws, coord, value)` require
  `value`. Turning the footgun into a compile error is the point; a doc note alone
  leaves it armed for the next report renderer. New `ensureCell(ws, row, col)` covers
  get-or-create.
- `freezePanes(ws, rows, cols)` removed; `setFreezePanes` now takes
  `'B2' | { rows, cols } | undefined`. Two exported ways to freeze panes was a
  "one way to do one thing" violation. The unified form is also strictly more capable:
  `freezePanes` required both counts `>= 1`, so "freeze two rows and no columns" was
  inexpressible.
- `WriteOnlyOptions.estimatedMaxRow` removed. Documented as reserved, never read.

**Added**

- `ensureCell` (`/worksheet`). The internal range helpers now use it instead of
  open-coding the same fallback four times.
- `registerCellStyle(wb, spec): number` (`/styles`). `setCell` already accepted a
  `styleId` and the stylesheet already deduped xf records, but there was no way to get
  an id without first having a cell; `setRangeStyle` built exactly that xf internally
  and threw the id away. `appendRow` / `appendRows` gained `{ styleIds }`, positionally
  aligned with the values. A column carrying an id is written even when its value is
  empty, which is how a bordered-but-blank input column survives an append.
- `patchCellFont(wb, c, patch)` (`/styles`). `setBold`, `setItalic`, `setStrikethrough`,
  `setFontSize`, `setFontName` and `setFontColor` are now this function with one field
  filled in, so there is a single merge path.
- `makeFormula(text, { cachedValue })` (`/cell`), and `setFullCalcOnLoad` is now
  exported from `/workbook` (it existed but was unreachable, which made the
  cached-value documentation unactionable).
- `SaveOptions.mtime` and `compressionLevel`, both wired through
  `createZipWriter`; `createWriteOnlyWorkbook` takes the same options.
- Range parameters across `/worksheet` and `/styles` accept
  `{ minRow, minCol, maxRow, maxCol }` as well as an A1 string (new `RangeRef` type in
  `/utils`). `writeRange` takes a `{ row, col }` anchor. `tupleToCoordinate` and
  `boundariesToRangeString` gained `absoluteCol` / `absoluteRow` for the `$B$5` and
  `$A$4:$H$20` forms formula text needs.

**Fixed**

- Every formula setter strips a leading `=`. OOXML stores `<f>` without it
  (ECMA-376 §18.3.1.40), so `setFormula(c, '=SUM(A1:A3)')` previously emitted
  `<f>=SUM(A1:A3)</f>` and Excel reported the file as damaged. `makeFormula` and
  `setFormula` had to agree on this or the pair would itself be a trap.
- `addExcelTable` rejects a table whose column count does not match the range width, or
  whose header cells do not hold their column names. Excel treats both as a damaged
  file and repairs it by dropping the table, so the failure surfaced as silent data
  loss far from the call. `headerRowCount: 0` skips the header check.

**Docs**

New recipes for styling by id, formulas in generated files (the three viewer classes
and what each does with an uncached `<v>`), asserting on a generated workbook, and
byte-identical output. Cheatsheet rows for the same, plus `setCellAsNumber`,
`autofitColumns`, `setColumnWidths` and an editable input column, all of which existed
and were not found. `llms.txt`, the README and the openpyxl migration guide updated to
match.

**Removed (internal, unreachable)**

`setCellFormula`, `setCellArrayFormula` and `setCellRichText` in
`src/worksheet/worksheet.ts` were dropped from the public surface in the
"one way per task" trim (`8ce7059`) and had been dead since. `setCell` plus
`makeFormula` / `makeRichText` covers them.

## Breaking changes

Three, all flagged `minor` in the changeset (pre-1.0) and listed under **Breaking**
above: mandatory `value` on `setCell` / `setCellByCoord`, `freezePanes` folded into
`setFreezePanes`, and `WriteOnlyOptions.estimatedMaxRow` dropped.

The first two are type-level, so affected code fails to compile rather than changing
behaviour silently. Migration is mechanical:

| Before | After |
| --- | --- |
| `setCell(ws, r, c)` | `ensureCell(ws, r, c)` |
| `setFormula(setCell(ws, r, c), text)` | `setCell(ws, r, c, makeFormula(text))` |
| `freezePanes(ws, 1, 1)` | `setFreezePanes(ws, { rows: 1, cols: 1 })` |

The `addExcelTable` validation is a behaviour change rather than a type change, and is
the one thing here that can turn working-looking code into a throw. That is the
intent: the files it now rejects are files Excel was already repairing by dropping the
table.

Where review effort is best spent: the `setCell` signature change (blast radius), the
`addExcelTable` validation (it can reject input that previously "worked"), and the
leading-`=` strip, which touches `setSharedFormula`'s optional-formula path.

## Checklist

      the result.
      marked accordingly.
      comments.


kibertoad:
Closing this in favour of six focused PRs. The work is unchanged; the packaging was wrong.

This PR bundled seven independent decisions behind one shared motivation, which made it un-reviewable as a unit and un-releasable in parts. A user who hit "Excel says your file is damaged" could not take the leading-`=` fix without also taking the `setCell` signature break, because one changeset covered both. And nobody could reject the range-parameter type unions without also blocking the ZIP determinism work, which shares no code with them.

Split, in suggested merge order:

| PR | What it decides | Breaking |
| --- | --- | --- |
| #136 | `makeFormula`, and strip the leading `=` OOXML does not allow in `<f>` | no |
| #137 | `SaveOptions.mtime` for reproducible bytes; wire `compressionLevel` | type-level |
| #138 | `addExcelTable` validates its columns and headers against the sheet | can newly throw |
| #139 | `ensureCell`, and `setCell`'s `value` becomes mandatory | yes |
| #140 | `registerCellStyle` and `patchCellFont` | no |
| #141 | Numeric coordinates where an A1 string was required; one freeze-panes API | yes |

#136, #137 and #138 are independent of each other and of the rest. #139, #140 and #141 all touch `src/worksheet/worksheet.ts` and `src/styles/cell-style.ts`, so whichever merges second will need a rebase; the conflicts are adjacent-line rather than semantic. The docs files (cheatsheet, `llms.txt`, the recipe registry) are touched by most of them and will conflict textually in the same way.

Each PR carries its own changeset, its own tests, and its own rationale for the design it picked and the alternatives it rejected. Reviewing them in any order is fine; the table above is just the order that minimises rebasing.

PR #136, opened 2026-09-16
feat: makeFormula, and strip the leading = that Excel rejects

kibertoad:
## Summary

Fixes a formula spelling that produced a workbook Excel calls damaged, and adds
`makeFormula` so placing a formula is one call instead of two. Both live in
`src/cell/cell.ts` and share one normalisation path, which is why they are one PR:
a value constructor that disagreed with `setFormula` about the leading `=` would be
a new trap rather than a fix for the old one.

## Motivation

Found while writing a workbook generator against `0.11.1`.

OOXML stores formula text without the leading `=` (ECMA-376 section 18.3.1.40), but
no setter stripped it. `setFormula(cell, '=SUM(A1:A3)')` emitted
`<f>=SUM(A1:A3)</f>`, and Excel reports that as an unreadable-content error on open.
The natural spelling was the broken one, and nothing in the types or the docs said
so.

Separately, placing a formula took two steps: reach or create a cell, then mutate it.
That is fine once, but a generated report does it for every formula cell and then
needs a number format on the same cell, so the two-step shape multiplies.

`setFullCalcOnLoad` turned out to exist but never be exported from
`@office-kit/xlsx/workbook`, which left the standard advice for generated files
("cache the value, or ask Excel to recalculate") with no reachable second half.

## Changes

- `setFormula`, `setArrayFormula`, `setSharedFormula` and `setDataTableFormula` strip
  a single leading `=`, so `'=SUM(A1:A3)'` and `'SUM(A1:A3)'` are interchangeable.
- `makeFormula(text, { cachedValue })` in `@office-kit/xlsx/cell` returns the
  `CellValue`, so `setCell(ws, row, col, makeFormula('SUM(B5:I5)'))` is the whole
  write and composes with the `styleId` argument `setCell` already accepted.
  `setFormula` stays as the mutate-a-cell-you-hold form.
- `setFullCalcOnLoad` is exported from `@office-kit/xlsx/workbook`.
- Docs: the formulas recipe now covers the three viewer classes and what each does
  with an uncached `<v>`; a new recipe shows asserting on a generated workbook;
  cheatsheet, `llms.txt` and the openpyxl migration guide follow.

Two changesets, so the fix and the addition get separate changelog lines and a user
can take the `patch` without the `minor`.

## Testing

`pnpm typecheck && pnpm lint && pnpm knip && pnpm test` and
`pnpm --dir site exec svelte-check` all pass. 2326 tests pass; the 36 failures on my
machine are the pre-existing `xmllint not available` conformance skips, identical in
count on `main`.

New assertions in `tests/cell/cell.test.ts` pin the strip across all four setters and
that `makeFormula` produces exactly what `setFormula` assigns. The strip test fails
without the change.

Reproducing the original bug:

```ts
const c = ensureCell(ws, 1, 1);
setFormula(c, '=SUM(A1:A3)');
// before: <f>=SUM(A1:A3)</f>, Excel reports unreadable content
```

## Breaking changes

None. Stripping a leading `=` only affects input that previously produced a file
Excel refused, and `<f>` content read back from a real workbook never carries one.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #137, opened 2026-09-16
feat!: pin the ZIP timestamp so identical input produces identical bytes

kibertoad:
## Summary

Two saves of the same workbook produced different bytes, because fflate stamps the
wall clock into every ZIP entry. `SaveOptions.mtime` pins it, which makes golden-file
tests and content-addressed caching possible for anything this library writes.
`compressionLevel`, which the same options bag already declared and never read, is
wired through in the same pass.

## Motivation

No issue. Hit while writing a determinism test for a workbook generator: the test had
to compare parsed cells rather than bytes, because `createZipWriter` constructs
`new ZipDeflate(path)` and never sets `mtime`, so fflate defaults each entry's local
header and central-directory record to `Date.now()`.

ZIP has no "no timestamp" encoding, so this cannot be fixed by omission; the
timestamp has to come from somewhere, and the caller is the only party that knows
what it should be. With `mtime` pinned and the core properties set from the caller's
own data rather than the clock, identical input produces identical bytes.

`SaveOptions` had exactly one field, `compressionLevel`, documented as "Reserved" and
never read. A declared option that silently does nothing is worse than no option, so
it is wired up here rather than left for later.

## Changes

- `SaveOptions.mtime?: Date`, applied to every entry by `createZipWriter`.
  `saveWorkbook` and `workbookToBytes` both accept it.
- `createWriteOnlyWorkbook` takes the same option, so the streaming path is
  reproducible too.
- `SaveOptions.compressionLevel` now reaches fflate's deflate stream, and is typed
  `0 | 1 | ... | 9` (the range fflate accepts) rather than `number`.
- `createZipWriter(sink, opts)` gained a second parameter, and `@office-kit/xlsx/zip`
  exports `ZipWriterOptions` and `CompressionLevel`. The latter has one canonical
  export, from `/zip`, following the public API convention.
- The stamp is encoded directly from UTC components into local-file and central-
  directory headers. This also handles daylight-saving gaps and date-line changes,
  which cannot be represented by adjusting a local `Date`. Entry payloads remain
  untouched, including stored payloads containing ZIP header signatures.
- Both options are validated in `createZipWriter` before the sink is opened: a year
  outside 1980-2099 (`new Date(0)`, the reflex reproducible-build stamp, is below
  it), an invalid `Date`, and a level off the `0..9` end all throw an
  `OpenXmlIoError` naming the option.
- Docs: a recipe on byte-identical output, plus a cheatsheet row and the `llms.txt`
  `SaveOptions` entry.

## Testing

`tests/io/deterministic-bytes.test.ts` covers buffered and streaming saves,
compression levels and round-trips, invalid options, and byte equality across
machine timezones. Additional regressions cover DST changes in New York and
Berlin, the skipped date in Apia, boundary years, two-second resolution, and both
local and central headers without modifying stored payloads. Five regressions
failed before the timestamp fix and pass afterwards.

Validation: all 2,456 tests pass, along with TypeScript, lint, Knip, build,
size limits, package export checks, TypeDoc, and the site's typecheck and build.

## Breaking changes

Two things change for a caller who already passes `compressionLevel`. The type
narrows from `number` to a `0..9` union, so `{ compressionLevel: someNumber }` no
longer typechecks. The value is also honoured now: their output used to come back at
fflate's default level 6 whatever they asked for, and it will change in size and in
bytes. A JavaScript caller passing something outside `0..9` now gets an
`OpenXmlIoError` where fflate used to substitute level 6 silently. A caller who
never set the option sees no runtime change. Flagged `minor` in the changeset, which
is how CLAUDE.md wants a pre-1.0 break flagged.

`WriteOnlyOptions.estimatedMaxRow` is also declared-and-ignored dead config, and it
stays that way here: a different interface in a different subpath, with nothing to do
with determinism. Touching it would make this PR's breaking surface wider than its
subject. Worth a one-line cleanup of its own if you want it.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #138, opened 2026-09-16
feat!: reject an Excel table that disagrees with the sheet under it

kibertoad:
## Summary

A table whose declared shape or column names disagreed with the sheet under it produced a
workbook Excel treats as damaged and repairs by dropping the table. Every path that
attaches a table to a worksheet now rejects one at the call.

## Motivation

No issue. Found while considering migrating a hand-drawn autofilter range to a real
Excel Table.

Excel does not report a disagreement as an error the author can see. It silently
"repairs" the file on open, so the delivered workbook loses its filter and its structured
references, and the symptom shows up in a customer's hands rather than in the generator.
That is the class of mistake this project's "validate at boundaries" rule exists for, and
both entry points already receive the worksheet, so the check costs nothing extra to
reach.

Three shapes of the fix I considered and did not take:

- **Check on save.** It would catch every ordering, including a header row overwritten
  after the table was added. It would also make `saveWorkbook` throw on a workbook
  `loadWorkbook` had just accepted, which breaks read-modify-write of a file another
  writer produced. Load and save stay lossless; the checks sit on the mutation path.
- **Fill a blank header row from `columns`.** Non-breaking, and what Excel's own "format
  as table" does. It also turns `addExcelTable` into a second way to write cell values,
  and it would have to invent the header styling. Writing the header row first is one
  line of docs.
- **`OpenXmlInvalidWorkbookError` for the throw.** The better fit on the class comments,
  but it is thrown nowhere in `src/`, while every builder-side validation here throws
  `OpenXmlSchemaError` (`setCellByCoord`, `mergeCells`, `freezePanes`, the defined-name
  helpers). Moving one of them alone is the inconsistency; moving all of them is a
  separate change.

## Changes

- `addTable` validates too, and `addExcelTable` goes through it, so there is one
  validating path onto `ws.tables` rather than one checked and one unchecked.
- A header cell pairs with a column name only when it holds text: a string, rich text, or
  a formula caching a string. Comparing through `cellValueAsString` made the verdict turn
  on a number format it never read (a header holding `42` under format `0.00` displays
  `42.00` and matched the name `"42"`), and a `Date` header could only ever match its
  ISO-8601 spelling.
- `ref` has to be a plain two-corner range. `rangeBoundaries` also accepts `A:C` and
  `1:5`, which fill the missing axis to the sheet maximum, so `ref: '1:1'` reached the
  width check as a 16384-column table, and `$` markers, which Excel never writes into
  `table/@ref` and which used to go through to the file verbatim.
- `ref` must fit `headerRowCount` + `totalsRowCount`; header-only tables are valid.
- Header and totals row counts must be unsigned 32-bit integers.
- Column names have to be non-empty and unique, compared case-insensitively because
  structured references are. `columns: ['', '']` on a sheet with no header row passed the
  old check (both sides rendered to `''`) and then wrote `<tableColumn name=""/>`, which
  this library's own reader rejects.
- The message separates an empty cell, an uncalculated formula, and a value Excel would
  not keep as header text, so the remedy it prints matches the cause. It names the table
  rather than the entry point, since two functions now raise it.
- Docs: `llms.txt` and the cheatsheet stop importing `addAutoFilter`, which
  `@office-kit/xlsx/worksheet` does not export, and stop offering `ws.tables.set(...)`,
  which `Worksheet.tables` has no method for.

## Testing

Maintainer review added regression coverage for header-only table round trips and invalid
row counts (negative, fractional, non-finite, or outside the unsigned 32-bit range). All
local tests, type checks, lint, build, size checks, and unused-code checks pass.

Original author validation:

`pnpm typecheck`, `pnpm lint`, `pnpm knip`, `pnpm build`, `pnpm size` and `pnpm test:e2e`
pass.

`pnpm test` is 2339 passed / 37 failed here. Every failing assertion is
`xmllint not available: spawnSync xmllint ENOENT`, the pre-existing XSD-runner gap on a
Windows box with no libxml2-utils; CI installs it. The OPC and semantic tiers are clean,
including the new conformance case.

Two tests earn a mention. The rejection of a numeric header is the one decision here that
reverses earlier behavior on this branch, so it is pinned by name and by message. And the
accepted case that replaces it, a rich-text header whose runs spell the column name, now
goes through `tests/conformance/writer-features.test.ts` rather than only asserting that
the builder returned two columns.

The snippets in `llms.txt` and the cheatsheet are string and markdown content that nothing
type-checks, which is how the `addAutoFilter` import survived in the first place. I
checked each name in them against `src/worksheet/index.ts` by hand.

## Breaking changes

`addTable` can now throw, and so can `addExcelTable` for a numeric or date header it used
to accept. Both are files Excel was already repairing by dropping the table, so the
alternative is shipping a workbook that has quietly lost a feature the author asked for.
Flagged `minor` in the changeset, per the pre-1.0 policy.

What a caller has to change:

- Adding the table before filling the header row: reorder the two statements, or pass
  `headerRowCount: 0`.
- A numeric or date header cell: write it as a string, which is what Excel stores in
  `tableColumn/@name` anyway.
- A whole-row or whole-column `ref`, or duplicate column names: use a finite cell range
  and unique non-empty names. Header-only tables remain supported.
- A `ref` spelled with `$` markers (`$A$1:$C$4`): drop them.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

Issue #111, opened 2026-09-02
bug: cells without a value are dropped on save, leaving defined names dangling


DywiTom:
## Summary

Cells that carry no value are not written back, so anything that references such a cell by address — a `definedName` in the attached file — is left pointing at a cell that no longer exists.

## Reproduction

Attached: `input.xlsx`, a minimal package that only contains what the bug
needs, and `output.xlsx`, what the script below turns it into.

```ts
import { fromArrayBuffer, loadWorkbook, workbookToBytes } from '@office-kit/xlsx/io';

// nothing is changed between load and save
const bytes = await Bun.file('input.xlsx').bytes();
await Bun.write('output.xlsx', await workbookToBytes(await loadWorkbook(fromArrayBuffer(bytes))));
```

```sh
unzip -p output.xlsx xl/worksheets/sheet1.xml
```

## Expected behavior

`xl/worksheets/sheet1.xml` comes back unchanged:

```xml
<row r="1"><c r="A1"><v>1</v></c><c r="B1"/><c r="C1" s="0"/></row>
```

## Actual behavior

```xml
<row r="1"><c r="A1"><v>1</v></c></row>
```

No error is raised; `loadWorkbook` and `workbookToBytes` both resolve normally.

## Environment

- @office-kit/xlsx version: 0.9.2
- Runtime: Bun 1.4.0 on Debian 12 x86_64
- Other relevant details: `input.xlsx` is a hand-written minimal package
  containing only what this bug needs — no Excel involved in producing it. We
  first hit the loss on real-world templates.

## Additional context

Dropping a genuinely unreferenced empty cell would be harmless. The problem is
that an empty cell can still be the target of

- a `definedName` (`Tabelle1!$B$1` in the attached file),
- a form control's `fmlaLink`,
- a drawing or comment anchor.

We hit this on an Excel-authored template that keeps an empty `<c r="C22"/>`
as the target of a defined name (`Kontrollkästchen2`, sitting under a check-box
control): after the round-trip the cell is gone and the name points into
nothing. Cells with a non-default style survive, so whether it bites depends on
the data.

[input.xlsx](https://github.com/user-attachments/files/31724883/input.xlsx)
[output.xlsx](https://github.com/user-attachments/files/31724884/output.xlsx)

## Confirmation

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I am on the latest released version, or I have explained above why an
      older version is relevant.
[...]

PR #139, opened 2026-09-16
feat!: ensureCell for get-or-create, and setCell's value becomes mandatory

kibertoad:
## Summary

`setCell(ws, row, col)` read like "reach the cell at (row, col)" and wrote `null`.
This makes `value` mandatory so the mistake is a compile error, and adds
`ensureCell(ws, row, col)` for what the three-argument form was actually being used
for.

## Motivation

A report generator built on `0.11.1` lost 34 of its 37 formulas to this: a
pass that stamped borders on eight columns per row ran over already-populated rows,
and every `setCell(ws, r, c)` it used to reach a cell blanked the formula that was
there. The author ended up carrying a local `cellAt = getCell(...) ?? setCell(...)`
helper.

Three things make this worth a breaking change rather than a docs fix:

- The docstring said "Create or update a Cell" and promised the styleId, hyperlinkId
  and commentId survive. It never said the value does not.
- The cheatsheet showed `setCell(ws, 3, 1, null)` and the formula recipe showed
  `setFormula(setCell(ws, 4, 1), ...)`, so the trap was the documented path.
- The library already had the get-or-create pattern internally, five times over:
  `applyToRange`, `setRangeStyle`, `setRangeWrapText`, `setRangeAlignment` and
  `setRangeBorderBox` each open-coded `ws.rows.get(r)?.get(c) ?? setCell(ws, r, c)`.
  It just was not public.

Documenting the semantics and leaving the signature alone was the other option. It
leaves the trap armed for the next caller who reads the docstring and reasonably
assumes the value survives, so this takes the pre-1.0 opportunity instead.

## Changes

- `ensureCell(ws, row, col)` in `@office-kit/xlsx/worksheet` returns the cell at a
  coordinate, allocating a blank one only when it does not exist. An existing cell is
  returned untouched. `ensureCellByCoord(ws, 'B5')` is the A1-addressed form, so that
  idiom keeps a reach-a-cell path now that `setCellByCoord` requires a value.
- `setCell` and `setCellByCoord` require `value`. To blank a cell deliberately, pass
  `null`.
- The five internal callers above call `ensureCell` instead of open-coding it, and the
  three `*ByCoord` wrappers share one strict coordinate parser instead of a copy of
  the regex each.
- Removed: `setCellFormula`, `setCellArrayFormula` and `setCellRichText` in
  `src/worksheet/worksheet.ts`. An earlier "one way per task" trim dropped them from
  the public subpaths and they have been unreachable since. `ensureCell` plus
  `setFormula` / `setArrayFormula` covers the formula pair. Rich text is
  `setCell(ws, r, c, { kind: 'rich-text', runs: makeRichText(runs) })`, because
  `makeRichText` returns the runs rather than a `CellValue`, unlike `makeErrorValue`
  and `makeDurationValue`.
- Docs: cheatsheet rows, the API reference cell-helper table, a recipe for a
  recipient-editable input column, and the openpyxl migration guide, which now
  contrasts `ws.cell(row=r, column=c)` against `setCell` explicitly, since openpyxl's
  no-value form is the non-destructive one.
- `llms.txt` is what an LLM reads as the API reference, and its array and shared
  formula snippets passed their arguments in the wrong order. That predates this PR
  and is fixed here because the same lines moved to `ensureCell`: `setArrayFormula`
  takes the ref before the formula, and `setSharedFormula` takes a numeric `si`, so
  the old snippet emitted `ref="TRANSPOSE(A1:A3)"` with `[object Object]` as the body
  and the shared one threw on `si`.

Where to look hardest: `mergeCells` prunes the cells under a merge, and `ensureCell`
on one of those coordinates allocates it again, so a loop over a range containing a
merge puts blank `<c>` elements back into `<sheetData>`. The range helpers have always
done that through their open-coded get-or-create, but `ensureCell` makes it reachable
in one call, so its docstring names the interaction and
`tests/worksheet/ensure-cell.test.ts` pins the on-wire result. Redirecting to the
top-left cell instead would change what `setRangeBorderBox` writes for merged ranges,
where Excel reads the block's borders off the covered cells.

## Testing

`pnpm typecheck`, `pnpm lint`, `pnpm knip`, `pnpm build`, `pnpm size`,
`pnpm check:attw`, `pnpm check:consumer` and `pnpm test:e2e` pass. `pnpm test` is 2325
passing; the 36 failures on my machine all come from the conformance XSD tier
reporting `xmllint not available`, which CI covers by installing `libxml2-utils`, and
none are in files this PR touches. `svelte-check` over `site/` flags only the
generated `api-data.json` that `pnpm doc:api` produces; the changed route and recipe
files are clean.

`tests/worksheet/ensure-cell.test.ts` pins the distinction that matters: `ensureCell`
returns an existing cell with its value and styleId intact, `setCell` replaces the
value, and a formula reached through `ensureCell` survives. It also pins the merge
case above and the `ensureCellByCoord` contract, including the coordinates it rejects.

It also covers deliberate blanking, since with the three-argument form gone
`setCell(..., null)` is the only spelling for it. The round-trip case is the one worth
reading: a blanked cell has to keep its fill through save and load, or the formatting
leaves with the value. That is a real risk here rather than a theoretical one, since
this repo has already had to fix valueless cells being dropped on save (#111).

The two doc-rot suites under `tests/smoke/` now assert and exercise `ensureCell`, so a
rename breaks CI rather than the README and the migration guide.

Most of the test diff is mechanical migration of the 18 call sites that used the
three-argument form, which is also the honest measure of the blast radius.

## Breaking changes

Yes, and this is the PR to scrutinise. `setCell` and `setCellByCoord` now require
`value`. It is type-level, so affected code fails to compile rather than changing
behaviour silently, and migration is mechanical:

| Before | After |
| --- | --- |
| `setCell(ws, r, c)` | `ensureCell(ws, r, c)` |
| `setCellByCoord(ws, 'B5')` | `ensureCellByCoord(ws, 'B5')` |
| `setFormula(setCell(ws, r, c), text)` | `setFormula(ensureCell(ws, r, c), text)` |

Emptying a cell stays available and is now explicit rather than implied. `null` is a
`CellValue`, so `setCell(ws, r, c, null)` is the "no value here" write: it clears the
value and leaves the cell in the sheet with its fill, border and number format, the
way Excel's Delete key does. `deleteCell(ws, r, c)` is the other meaning and drops the
cell outright, formatting included; `clearRange` does that across a rectangle.

Flagged `minor` in the changeset, per the pre-1.0 convention in this repo.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

baseballyama (approved):
Thank you so much!

PR #140, opened 2026-09-16
feat: registerCellStyle and patchCellFont, so styling is not a second pass

kibertoad:
## Summary

Adds `registerCellStyle`, which turns a style spec into a `styleId` that `setCell`
takes as its last argument and `appendRow` / `appendRows` take inside
`{ styleIds: [...] }`, so formatting arrives with the value instead of in a second
pass over every cell. Adds `patchCellFont` for the merge case, and rebuilds the seven
single-field font setters on top of it.

## Motivation

No tracking issue. Both gaps turned up while writing a report generator against
`0.11.1`, so there is no earlier bug report to link.

**Styling cost too many calls.** `setCell` already accepted a `styleId` and the
stylesheet already deduped xf records, but nothing could turn a style spec into an id
without a cell to hang it on first. Without it, a 16-row sheet ran one loop to stamp
`#,##0` on three columns and a second to stamp a border on eight, roughly 200 style
calls for 16 rows, and the second pass over already-populated cells is what makes this
dangerous rather than merely slow.

**Replacing a font silently drops the workbook default.**
`setCellFont(wb, c, makeFont({ bold: true }))` is a legal call that registers
`<font><b/></font>` with no `<sz>` and no `<name>`. Excel substitutes its own
default, LibreOffice and Sheets may render differently, and nothing warns. The
generator ended up wrapping every font change in a local helper that spread
`getCellFont` before `makeFont`. The single-field setters (`setBold`, `setFontSize`,
and friends) already did that merge internally, one field at a time, so "bold and
size 13" still cost two calls and two intermediate xf records.

## Changes

- Saving throws `OpenXmlSchemaError` when a cell's `styleId` names no entry in its
  workbook's `cellXfs` pool, naming the sheet and the cell. Most of the risk in this
  PR sits here: a `styleId` is an index into one workbook's stylesheet, and handing
  callers an opaque number makes "id from workbook A, cell in workbook B" reachable.
  Excel opens the resulting file with the repair dialog and drops the sheet, so the
  write fails instead. `styleId 0` stays legal on an empty pool, since it emits no
  `s=` attribute at all.
- `registerCellStyle(wb, spec): number` in `@office-kit/xlsx/styles`. The id names a
  complete style rather than a patch over the target cell, so an axis missing from
  the spec renders as the workbook default. `setCellStyle` and `setRangeStyle` remain
  the patch-an-existing-cell paths.
- `appendRow` and `appendRows` take `{ styleIds }`, positionally aligned with the
  values and reused for every row of an `appendRows` call. A column with a style id
  is written even when its value is empty, so a bordered-but-blank input column
  survives the append. Ids past a row's last value therefore add styled blanks and
  widen the sheet, which the docs and a test now spell out along with the per-row
  `slice` opt-out.
- `patchCellFont(wb, cell, patch)` merges a partial font over the cell's current one.
  A field set to `undefined` is removed rather than ignored, which is what lets
  `setUnderline` join `setBold`, `setItalic`, `setStrikethrough`, `setFontSize`,
  `setFontName` and `setFontColor` on the single merge path instead of hand-rolling
  its own removal.
- `makeFont` and `patchCellFont` take a new exported `FontPatch` in place of
  `Partial<Font>`. Under `exactOptionalPropertyTypes` the latter cannot express
  `{ size: undefined }`, so clearing a field was unreachable through the type even
  though the runtime already did the right thing.
- `CellStyleSpec` is exported from `@office-kit/xlsx/styles` and lives in
  `styles/stylesheet.ts`, next to the one spec-to-xf builder that
  `registerCellStyle`, `setCellStyle`, `setRangeStyle` and the streaming write-only
  writer all call. `WriteOnlyStyle` is an alias of it. The builder cannot live in
  `cell-style.ts`: that module imports `worksheet.js`, which would land in the
  `/streaming` dependency graph and its size budget.
- `setCellFont` and `setRangeFont` docstrings now say they replace rather than merge,
  and point at the merge path.
- Docs: a recipe for styling a report by id, notes on the replace/merge distinction
  in the existing styling recipe, cheatsheet rows, `llms.txt` and the openpyxl
  migration guide.

## Testing

`pnpm typecheck && pnpm lint && pnpm knip && pnpm test && pnpm build && pnpm size &&
pnpm check:attw && pnpm check:consumer`, plus `pnpm --dir site exec svelte-check`.
2339 tests pass. The 36 failures on my machine are the pre-existing `xmllint not
available` conformance skips, identical in count on the base commit without this
diff. `/streaming` comes in at 203.25 kB against its 220 kB budget.

`tests/styles/register-cell-style.test.ts` pins the claims a reviewer would otherwise
have to take on trust: that an id from another workbook fails the save rather than
reaching the file, that equal specs dedup to one xf, that an id is a whole style
rather than a patch, that an omitted axis resolves to pool slot 0 even when a loaded
workbook put something non-neutral there, and that ids outrunning a row's values
widen the sheet. `tests/styles/patch-cell-font.test.ts` pins that an explicit
`undefined` clears a field while an omitted key keeps it.

## Breaking changes

One behavior change, which I do not think rises to breaking. `saveWorkbook` and
`workbookToBytes` now throw `OpenXmlSchemaError` for a cell whose `styleId` falls
outside its workbook's `cellXfs` pool. Reaching that state takes either a raw
`setCell(..., n)` with an out-of-range `n` or an id carried between workbooks, and
the file it used to produce was one Excel refused to open without repairing away the
sheet. Nothing that produced a working file before stops working.

`registerCellStyle`, `patchCellFont`, `CellStyleSpec`, `FontPatch` and the `styleIds`
option are additions. `FontPatch` on `makeFont` is a widening of an existing
parameter, and `WriteOnlyStyle` keeps its name and its shape. The single-field
setters keep their behaviour: they merged before and they merge now, through one
shared path instead of seven copies.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #141, opened 2026-09-16
feat!: accept numeric coordinates where an A1 string was required

kibertoad:
## Summary

Range-taking helpers now accept pre-computed bounds as well as an A1 string, validating
them the way the A1 parser does, and the coordinate helpers can emit the `$` forms
formula text needs. Also folds `freezePanes(ws, rows, cols)` into `setFreezePanes`, so
there is one way to freeze panes rather than two.

## Motivation

Found while writing a report generator on `0.11.1` that tracks rows and
columns as integers.

`setRangeStyle`, `setRangeNumberFormat`, `setRangeBorderBox`, `applyToRange`,
`writeRange` and the rest all insisted on an A1 string, so the generator imported
`columnLetterFromIndex` and formatted `"A4:H20"` for the callee to parse straight
back into the numbers it started from. The parse step already produces
`CellRangeBoundaries`, so accepting them is a type union plus an early return.

Formula text has the same shape of problem, and no helper emitted the `$` forms, so
`$B$5` and `${totalLetter}${firstDataRow + index}` came out of string concatenation.

Separately, `setFreezePanes(ws, 'A5')` and `freezePanes(ws, rows, cols)` were both
exported and did the same job. That is the parallel path this project's
"one way to do one thing" rule rules out, and the choice is not arbitrary: the
numeric form is strictly more capable once folded in, because `freezePanes` required
both counts to be at least 1, so "freeze two rows and no columns" could not be
expressed at all.

## Changes

- New `RangeRef` type in `@office-kit/xlsx/utils`:
  `string | CellRangeBoundaries`. Every range parameter across
  `@office-kit/xlsx/worksheet` and `@office-kit/xlsx/styles` takes it, `copyRange` /
  `moveRange` / `mergeCells` / `unmergeCells` included. `setSelectedRange` keeps its
  string, because an `sqref` holds several ranges, and the `*Str` helpers stay
  string-in / string-out by definition.
- Numeric bounds resolve through `makeCellRange`, so they are validated against the
  grid and inverted bounds are normalised. `{ minRow: 5, maxRow: 1 }` covers the same
  cells as `"A5:A1"`, while row 0, a fractional edge, `NaN` and `maxRow: 1e15` throw
  rather than iterating a region Excel cannot express.
- `writeRange` takes a `{ row, col }` anchor as well as an A1 ref, grid-checked before
  the first write rather than on the first non-null value.
- `mergeCells` registers a normalised copy, so mutating a caller-held bounds object
  cannot rewrite a merge already on the sheet, and `getRangeAddress` normalises too, so
  the address it returns names the region the other helpers operate on.
- `tupleToCoordinate` and `boundariesToRangeString` take
  `{ absoluteCol, absoluteRow }`, for `$B$5` and `$A$4:$H$20`.
- `setFreezePanes` accepts `'B2' | FreezeCounts | undefined`; `freezePanes` is
  removed. `freezeRows` / `freezeColumns` compose the counts form instead of assembling
  a ref of their own, and "nothing to freeze" is rejected once, by `makeFreezePane`,
  whichever spelling arrives.
- Docs: cheatsheet rows and snippets, the `llms.txt` freeze block and export table, and
  the merge-and-freeze recipe, which was still teaching
  `makeSheetView({ pane: makeFreezePane(...) })`.

A new `cellRef(row, col, opts)` helper was the other way to cover the formula case.
It would sit next to `tupleToCoordinate` doing the same job with the arguments
swapped, which is the parallel path again, so the existing helper grew the options
instead.

Widening one parameter rather than adding a second function per capability is the
deliberate call: each helper stays the single canonical path and reads either spelling,
the same shape `setFreezePanes` now has. That reasoning sits at the `RangeRef`
definition, so the next reader meets it where they meet the type.

## Testing

`pnpm typecheck && pnpm lint && pnpm knip && pnpm test`, then
`pnpm build && pnpm size && pnpm check:attw && pnpm check:consumer`, and
`pnpm --filter @office-kit/xlsx-site check` (491 files, 0 errors) all pass. 2345 tests
pass; the 36 failures on my machine are the pre-existing `xmllint not available`
conformance skips, identical in count on `main`.

`tests/worksheet/numeric-range-refs.test.ts` asserts the numeric and A1 forms produce
identical results on the same sheet, rather than just that the numeric form runs. It
also pins the bounds the numeric path now rejects, that `setRangeValues` stays inside
its rectangle, and that a registered merge survives mutation of the object it was built
from. `tests/utils/absolute-refs.test.ts` covers each `$` combination including past the
single-letter columns.

## Breaking changes

`setRangeValues` clips to the range it is handed. Values past the bottom or right edge
are dropped instead of written outside it, which is what `copyRange` already does
against a smaller target, and it makes `getRangeValues` the inverse its docstring
claims. This is the one change here that fails quietly rather than at compile time:
`setRangeValues(ws, 'A1', rows)` used to lay a whole block down from a one-cell range
and now writes one cell. `writeRange(ws, 'A1', rows)` is that behaviour, and it returns
the bounding box it wrote. Nothing in the repo relied on the overflow.

`freezePanes(ws, rows, cols)` is removed. Type-level, so affected code fails to
compile; replace `freezePanes(ws, 1, 1)` with
`setFreezePanes(ws, { rows: 1, cols: 1 })`.

Both are flagged `minor` in the changeset, per the pre-1.0 convention in this repo.
Widening the range parameters is otherwise additive, and existing A1 callers are
unaffected.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #144, opened 2026-09-17
perf: cap the SAX feed size so sheet walks stop buffering a chunk's worth of events

kibertoad:
## Summary

Walking a worksheet queued every SAX event a single parser write produced before
yielding the first row, so peak heap tracked whatever the parser was handed rather than
staying flat. On a 200k-row sheet, `iterRows({ minRow })` above row 1 goes from about
1240 MB peak heap to about 70 MB, and a default full-sheet walk from about 200 MB to
about 60 MB. Fixing the band path also surfaced that it was broken outright on sheets
Excel wrote, which this PR fixes too.

## Motivation

`iterParse` feeds saxes, and saxes runs its handlers synchronously inside `write()`.
Everything one `write()` produces is queued before the consumer sees any of it, so the
size of the argument sets the size of the queue.

Two different inputs were oversized, for two different reasons:

- Bytes and strings were fed as one `write()` of the whole document. A 40 MB sheet body
  peaked around 990 MB. `iterRows({ minRow })` above row 1 hits this, because that path
  hands the parser a materialised buffer, and skipping a header row is the ordinary thing
  to do with a streamed sheet.
- A `ReadableStream` was fed one upstream chunk per `write()`, and that chunk size is the
  producer's choice rather than ours. The zip reader pushes 64 KB of compressed bytes per
  pull and worksheet XML inflates about sevenfold, so a real sheet arrives in chunks
  around 450 KB. Measured on the 200k-row fixture: 80 chunks, mean 457 KB, max 471 KB.
  This is the path the default full-sheet walk uses.

There is no issue for this. It came out of a performance audit of the read and write
paths; the remaining items from that audit are being opened separately.

## Changes

- `iterParse` caps what it feeds saxes at 64 KB for every input shape. For a stream the
  cap is applied to the bytes before they are decoded, so a producer that hands over a
  whole part in one chunk no longer materialises the document as a string either.
- `iterRows({ minRow })` above row 1 works on sheets Excel wrote. The band was replayed
  inside a rebuilt `<sheetData>` element that declared only the default namespace, so the
  `x14ac:dyDescent` Excel puts on nearly every `<row>` raised an unbound-prefix error and
  the query threw. The replay now reuses the part's own bytes up to the end of
  `<worksheet …>` as its header, so whatever the file declared stays in scope.
- A malformed document throws `OpenXmlSchemaError` with the parser's error as `cause`,
  rather than surfacing the raw `saxes` `Error`.
- `iterParse` cancels the stream it was given when the consumer stops early. Abandoning
  the iteration is routine (`iterRows({ maxRow })` returns the moment the band ends), and
  the zip reader frees its inflate state from `cancel()`.
- The DTD prescan runs per chunk with an overlap carry rather than once over the prologue.
  Behavior change worth calling out: the `ReadableStream` path previously scanned only the
  first 256 characters and now scans the whole document, so a `<!DOCTYPE` or `<!ENTITY`
  token further in, inside a comment or a CDATA section for instance, is now rejected
  where it used to parse.
- Band queries replay their slice as views over the sheet bytes instead of concatenating a
  copy of the remainder.
- The row-offset index reads `r="N"` directly from the bytes instead of decoding each
  row's attribute region and running a regex over it.

## Review corrections

Band replay preserves the original prefix through the opening `sheetData` tag,
including namespace declarations on that element and processing instructions.
The source stream lock is released after cancellation. Two regressions failed
before these fixes and pass afterwards. All 2,466 default tests pass.

## Testing

Where the risk sits, and what pins it:

- The band path was broken on real workbooks and no test caught it, because every fixture
  in `tests/streaming/` is written by this library and our writer emits no prefixed
  attributes on `<row>`. The new case in `tests/streaming/row-offset-index.test.ts` runs
  against an Excel-emitted fixture from the openpyxl corpus and asserts the band equals
  the full walk filtered to the same rows. On `main` it throws
  `unbound namespace prefix: "x14ac"`.
- The regression tests for the chunking itself put a mismatched close tag after a body
  larger than the feed size, one per input shape: fed in one `write()` the throw lands
  before any event escapes and the consumer sees zero, fed in chunks the leading elements
  come out first. Both assert `0 > 0` and fail without the change.
- Comparing string, byte and stream input against each other only proves the three agree,
  and they all now run through the same chunker. A separate case compares a document
  larger than one chunk against a hand-derived event list, joining adjacent text events
  first, since a chunk boundary inside a text node legitimately splits it into two.
- Stream cancellation is asserted through a source whose `cancel()` sets a flag, with the
  consumer breaking after three events.

`tests/perf/iterparse-heap.test.ts` asserts a peak-heap ceiling per input shape. It moves
from `PERF_HEAP_GATE` to `PERF_GATE`, which is what the CI perf job sets, so the ceiling
now actually runs there; `heap.test.ts` keeps `PERF_HEAP_GATE` because its cells-per-MB
floor sits close enough to the observed value to flap on a shared runner. The perf config
pins `--expose-gc`: without it the baseline includes the fixture builder's garbage, and
V8 reclaiming that mid-walk can drive the measured delta to zero and pass the assertion
whatever the parser does.

```sh
pnpm test tests/xml/iterparse.test.ts tests/streaming/
PERF_GATE=1 pnpm test:perf tests/perf/iterparse-heap.test.ts
```

Measured on Node 24.18, Windows, 200k rows x 5 cells. Peak `heapUsed` sampled during
iteration, after a forced collection, lower is better. Both columns come from the same
harness; `before` is `main`.

`iterParse` over a 40 MB sheet body, 5.4M events:

| input shape | before | after |
| --- | --- | --- |
| `Uint8Array` | 987 MB, 2921 ms | 64 MB, 2402 ms |
| `string` | 947 MB, 2861 ms | 64 MB, 2477 ms |
| `ReadableStream`, 64 KB chunks | 64 MB, 2609 ms | 64 MB, 2404 ms |
| `ReadableStream`, 457 KB chunks | 382 MB, 3042 ms | 63 MB, 2451 ms |

`loadWorkbookStream` then `iterRows`, 3.8 MB archive:

| `minRow` | before | after |
| --- | --- | --- |
| 1 | 202 MB, 3803 ms | 62 MB, 3502 ms |
| 2 | 1239 MB, 4570 ms | 69 MB, 3457 ms |
| 100 000 | 660 MB, 2329 ms | 73 MB, 1767 ms |
| 199 000 | 0 MB, 265 ms | 0 MB, 220 ms |

What remains on the full-sheet walk is not queued events: the inflate allocates a fresh
450 KB buffer per chunk, and the 200k row arrays and 1M cell objects are what the caller
asked for. Figures move by tens of MB between runs on the same code, so treat them as
orders of magnitude rather than measurements to regress against; the gate in
`tests/perf/` is where the ceiling lives.

Row-attribute read in isolation, 1M rows: 160 ms decoding plus regex, 10 ms scanning
bytes. It does not move the end-to-end figures above, which the SAX walk dominates.

## Breaking changes

None to the public API surface. Two behavior changes a caller can observe, both in the
changeset:

- A malformed document now throws `OpenXmlSchemaError` rather than the raw `saxes`
  `Error`. Code catching `Error` is unaffected; code matching on the concrete class or on
  the unprefixed message is not.
- Streamed input is prescanned for DTD and entity declarations over the whole document
  instead of its first 256 characters, so a payload carrying one of those tokens past the
  prologue is now rejected. This closes a gap rather than adding a restriction: the same
  input already threw when it arrived as bytes or a string.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #145, opened 2026-09-17
perf(io): stream worksheet parts into the zip instead of buffering them

kibertoad:
## Summary

Saving a workbook held every worksheet's XML in memory at once. A sheet was built as one
string fragment per cell, joined, and encoded, and `saveWorkbook` then kept every sheet's
encoded bytes until the last one was serialised. Sheets now serialise straight into their
ZIP entry through a 64 KB flush buffer, so the heap a save needs no longer scales with the
sheet.

## Motivation

There is no issue for this. It came out of a performance audit of the read and write paths,
and this was the largest finding on the write side. The remaining items from that audit are
being opened separately.

The ZIP writer underneath already streams: `addEntry` pushes bytes through fflate's deflate
stream as the call runs, and the file sinks forward chunks to disk without holding the
archive. The save path was the layer negating that. For a 500k-cell sheet producing 3 MB of
output, the fragment array, the joined string and the encoded bytes were all live at once.

Measured as the smallest `--max-old-space-size` under which the save completes, against the
built bundles on Node 24.18, Windows. Lower is better.

| workload | before | after |
| --- | --- | --- |
| 1 sheet, 500k cells, 3.0 MB archive | 192 MB | 88 MB |
| 8 sheets, 2M cells, 9.8 MB archive | 384 MB | 320 MB |

The multi-sheet case improves less because the `Workbook` model itself dominates there.
This change does not touch the model.

Retention during the save, sampled after a forced GC from inside the sink while the eighth
sheet is being written, which is what `tests/perf/save-heap.test.ts` gates on: **80.3 MB
before, 5.4 MB after**, identical across three runs each.

One note on methodology, because the obvious metric points the wrong way. A sampled
`heapUsed` peak reports this change as a regression, 1.63x of part size before against
2.56x after. Streaming swaps one large retained buffer for a stream of short-lived
fragments, so peak *allocation* rises while the memory *ceiling* falls. Retention and the
heap-cap sweep both measure the ceiling, which is what decides whether a save fits in a
container.

## Changes

- `writeWorksheetXml(ws, ctx, emit)` in `src/worksheet/writer.ts` serialises a worksheet
  into a sink one XML fragment at a time. `worksheetToBytes` stays as the collect-and-join
  wrapper for callers that want the whole part in hand.
- `saveWorkbook` settles part names and workbook rIds in a cheap first pass over
  `wb.sheets`, writes `xl/workbook.xml`, then serialises each sheet directly into its ZIP
  entry and releases it before starting the next.
- `xl/_rels/workbook.xml.rels` moves after the worksheet parts in the archive. Whether it
  needs a `sharedStrings` relationship is only known once every sheet has been serialised,
  since that is what fills the shared-strings table.
- A save that fails part-way through now clears the destination before it rejects. Sheets
  reach the sink as they serialise, so a serialisation error can arrive with bytes already
  on disk, and `toFile`'s cleanup raced its own file-handle close and lost on Windows. The
  abort chain is awaited end to end and waits for the handle.

No public API change. `writeWorksheetXml` is internal, like the `worksheetToBytes` it sits
beside; neither is re-exported from a subpath.

## Testing

`tests/io/save-streaming.test.ts` runs in the default suite and covers the fragment-level
contract, that streamed output is byte-identical to `worksheetToBytes`, the part ordering,
and that the `sharedStrings` relationship survives being discovered mid-serialisation. That
last one is the case the reordering exists to protect: get it wrong and Excel opens the file
with every string cell blank.

`tests/perf/save-heap.test.ts` gates retention at 20 MB, roughly 4x clear of the 5.4 MB
reading and 4x under the 80.3 MB it would be without this change. CI runs it under
`PERF_GATE=1 pnpm test:perf`.

`vitest.perf.config.ts` passes `--expose-gc` to the perf workers. Vitest spawns them with
its own execArgv, so without it `globalThis.gc` is undefined, the forced GC does nothing and
the gate has no way to tell retention from garbage.

The round-trip and conformance fixtures are the real safety net for a save restructure, and
they pass unchanged, pivot tables and `.xlsm` included.

Reproduce the heap-cap figures:

```sh
pnpm build
cat > /tmp/cap.mjs <<'EOF'
import { createWorkbook, addWorksheet } from './dist/workbook.mjs';
import { appendRow } from './dist/worksheet.mjs';
import { workbookToBytes } from './dist/io.mjs';
const wb = createWorkbook(), ws = addWorksheet(wb, 'S');
for (let r = 0; r < 100_000; r++) appendRow(ws, [r, `row-${r}`, r * 1.5, true, r % 7]);
console.log((await workbookToBytes(wb)).length);
EOF
node --max-old-space-size=88 /tmp/cap.mjs    # passes here, fails on main
node --max-old-space-size=192 /tmp/cap.mjs   # smallest cap main survives
```

## Breaking changes

`xl/_rels/workbook.xml.rels` moves after the worksheet parts, so the bytes of every saved
archive change. Readers are unaffected, since `loadWorkbook`, zip libraries and Excel all
resolve parts by name. Byte-for-byte consumers are not: anyone who pinned `mtime` for reproducible
output in 0.13.0 and stores golden files or content hashes of saved workbooks has to
regenerate them, so the changeset is marked `minor`.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #146, opened 2026-09-17
perf: setHyperlinks and setComments, so a link or note on every row is not quadratic

kibertoad:
## Summary

Setting one hyperlink or comment scans the sheet's list to find the ref it replaces, so a
sheet that carries one per row costs time quadratic in the number of rows. `setHyperlinks`
and `setComments` take the whole batch and resolve it against one index, built and dropped
inside the call: 16 000 hyperlinks go from about 380 ms to about 7 ms, 16 000 comments from
about 270 ms to about 3 ms. Each batch leaves the sheet exactly as the matching run of
single calls would.

The single-entry functions are untouched. No index outlives a call, so `ws.hyperlinks`,
`ws.legacyComments` and the entries in them stay ordinary mutable objects, correct to read
and edit from anywhere.

## Motivation

`setHyperlink` calls `removeHyperlink` before appending, and `removeHyperlink` is a
`findIndex` over the full list. Adding N distinct refs makes every one of those calls miss
after walking everything added so far, which is the quadratic term. `setComment` has the
same shape.

A report generator that puts a link or a note on each row hits this directly, and it
degrades silently: nothing fails, the build just gets slower the bigger the sheet gets.

A persistent index behind the single-entry functions would be faster still, and it cannot
be made correct here: the arrays and the entries in them
are public and mutable, so assigning into an occupied slot, or setting `ref` on an entry the
cache already holds, leaves the array's length unchanged and the cache wrong. Noticing
either costs a scan of the array, which is the cost the cache exists to remove. A batch
sidesteps the question by not existing between calls.

There is no issue for this. It came out of a performance audit of the read and write paths;
the remaining items from that audit are being opened separately.

## Changes

- `setHyperlinks(ws, entries)` and `setComments(ws, entries)`, exported from
  `@office-kit/xlsx/worksheet`. Each applies its entries in order with the result of the
  matching single calls, including the orderings those differ on: a replaced hyperlink moves
  to the end, a replaced comment stays where it sits.
- `setHyperlinks` runs the same `target` validation as `setHyperlink` and applies nothing
  until every entry has passed, so a bad entry leaves the sheet as it was. The message names
  the offending index and ref.
- A part carrying the characters `<!DOCTYPE` or `<!ENTITY` as content, in a comment or a
  CDATA section, now loads. Chunked feeding had turned the prescan that rejects DTDs into a
  whole-document text search, so those files were rejected. The scan now stops where a
  declaration stops being legal, at the root element, and the patterns match the whitespace
  the grammar requires after either keyword, so a chunk that ends on a bare `<!ENTITY` is no
  longer mistaken for one.

A plural setter next to a singular one is a second path to the same result, which this repo
does not add lightly. The reason it earns its place is the cost curve rather than the
ergonomics: the batch is linear where the loop is quadratic, and no amount of care inside
`setHyperlink` can fix a loop it cannot see. `appendRow` / `appendRows` and `unhideRow` /
`unhideRows` already stand in that relation here.

Left out deliberately: `getCellHyperlink` and `getCellComment` still walk every entry and
parse its `ref` per cell, so resolving links for a whole rendered sheet stays O(cells x
entries). A `ref` there may be a range (`A1:B5`), so a ref-keyed index cannot serve them and
a coordinate index is a separate change with its own measurement.

## Testing

`tests/worksheet/ref-batch.property.test.ts` generates batches over seeded arrays, including
duplicate refs pushed on directly, and asserts the sheet ends up identical to the same
sequence applied one call at a time. That equivalence is the whole claim of the API, so it
is worth generating rather than enumerating.

`tests/worksheet/ref-batch.test.ts` pins the two orderings by example, that the array keeps
its identity so a held reference still sees the result, that an invalid entry writes
nothing, and the algorithmic change itself: with 5 000 entries already on the sheet it
counts `ref` property reads rather than wall-clock, so it cannot flake.

The parser cases pin a document whose body quotes `<!DOCTYPE` past the first feed chunk, and
a chunk that ends on the bare `<!ENTITY` keyword with the next one ruling the declaration
out.

```sh
pnpm test tests/worksheet/ref-batch.property.test.ts tests/worksheet/ref-batch.test.ts tests/xml/iterparse.test.ts
```

Measured on Node 24, Windows, one entry per row on an otherwise empty sheet. Lower is
better.

| entries | one at a time | batch |
| --- | --- | --- |
| 4 000 hyperlinks | 25 ms | 2 to 3 ms |
| 16 000 hyperlinks | 372 to 399 ms | 7 to 8 ms |
| 4 000 comments | 24 to 26 ms | 1 ms |
| 16 000 comments | 265 to 283 ms | 3 ms |

`@office-kit/xlsx/worksheet` grows from 10.44 kB to 10.71 kB brotlied, against a 100 kB
budget.

## Breaking changes

None. The one behaviour change is in the permissive direction: a file that was rejected for
carrying `<!DOCTYPE` as content now loads. Nothing that loaded before stops loading, and a
real declaration in the prologue is still rejected.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (changes requested):
Two correctness regressions block this cache design because both arrays and their entries are publicly mutable:

1. After `setHyperlink(ws, 'A1', { target: 'https://a.example' })`, replace `ws.hyperlinks[0]` with `makeHyperlink({ ref: 'Z9', target: 'https://z.example' })`. The first `getHyperlink(ws, 'Z9')` returns `undefined`. Looking up the old ref first happens to rebuild the index and hides this bug in the current test.
2. Add comments at A1 and A2, get the first comment, and change its `ref` to A2. `getComment(ws, 'A2')` returns the second comment, and `removeComment` removes the wrong entry instead of the first duplicate. The cached hit still matches, so the hit check cannot detect this.

Both cases fail against this head and preserve the same array length. Rebuilding only on misses fixes the first case but leaves the second case and makes distinct inserts quadratic again. Please preserve the mutable public model contract; a cache needs explicit invalidation or an opt-in/batch API with a defined lifetime before it can safely deliver this optimization. A transparent persistent index over arbitrary mutable arrays cannot infer these edits from length and a single cached hit.

This branch also includes an older version of #144; please base the revised implementation on current main so the namespace, parser cleanup, and streaming fixes from the merged #144 are retained.

kibertoad:
You were right, and the fix is a different shape rather than a patch: there is no cache between calls any more.

Both cases you listed come from the same root. The arrays and the entries in them are public and mutable, so neither an assignment into an occupied slot nor a `ref` set on an entry the cache already holds changes the array's length. I could not find a way to notice either for less than the scan the cache exists to remove, so the persistent index is gone.

What replaces it is `setHyperlinks(ws, entries)` and `setComments(ws, entries)` in 885f00c: one index, built inside the call and dropped with it. The single-entry functions are untouched, so `getHyperlink`, `removeHyperlink`, `setComment` and direct edits to `ws.hyperlinks` keep exactly the contract they have on main. 16 000 hyperlinks go from about 380 ms to about 7 ms, 16 000 comments from about 270 ms to about 3 ms.

The claim that deserves the scrutiny now is equivalence, so it is generated rather than enumerated: `tests/worksheet/ref-batch.property.test.ts` applies a random batch over a seeded array, duplicate refs included, and asserts the sheet ends up identical to the same sequence applied one call at a time. That covers the two orderings the single calls differ on, a replaced hyperlink moving to the end and a replaced comment staying where it sits.

Rebased onto current main, so #144, #145 and #147 are all in. Of the parser work that was sitting on the old head, main already covers the chunk copy and the row-band replay, so those commits are dropped. One fix is left, 69bebea: the prescan reads the whole document, so a part carrying `<!DOCTYPE` or `<!ENTITY` as content in a comment or a CDATA section is rejected rather than read, and because `\b` also matches at end of string, a chunk that ends on the bare keyword is rejected too. It now stops at the root element, where a declaration stops being legal, and matches the whitespace the grammar requires after the keyword. Say the word and I will move it to its own PR so this one stays on the batch API.

One more thing worth its own look rather than this PR: `tests/perf/iterparse-heap.test.ts` samples `heapUsed` without collecting first, so the figure is garbage produced since V8 last collected rather than what the walk retains. On the run of my old head the stream case measured +272 MB on the ubuntu runner and +0 MB on my laptop for the same code, against the 150 MB ceiling, so the gate passes or fails on GC mood. Collecting at each sample puts both shapes at about 2 MB, which a single tight ceiling can cover. I can open that separately if you want it.

src/xml/iterparse.ts
```diff
@@ -231,7 +240,11 @@ export async function* iterParse(input: SaxInput): AsyncIterableIterator<SaxEven
   };
 
   for await (const chunk of decodedChunks(input)) {
-    scanForDtd(chunk);
+    // The prologue is the only place a declaration is legal, and it is fed
+    // before the root element opens. Past that, `<!DOCTYPE` in the text is a
+    // literal inside a comment, CDATA or character data, and scanning on would
+    // both reject a valid sheet and re-read the whole document for nothing.
+    if (!rootOpened) scanForDtd(chunk);
```
baseballyama: [P2] Stop the prescan at the actual root boundary, including within this chunk. `rootOpened` is only updated by the subsequent `feed(chunk)`, so the entire first chunk is still scanned. For example, `<r><!-- <!DOCTYPE html> --><t>ok</t></r>` throws when supplied as a string, bytes, or one stream chunk, but succeeds when the identical stream is split into `<r>` and the rest. CDATA containing `<!ENTITY foo>` behaves the same way. Thus the advertised valid-content fix still fails for small XML parts, and acceptance now depends on upstream chunk boundaries. Please make the prologue scan respect XML lexical context and the root boundary within a chunk, with equivalent-input tests for all input shapes while retaining real-DTD rejection.

src/worksheet/ref-batch.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  // A removal tombstones its slot rather than splicing, which is what keeps the
+  // pass linear. Survivors keep their relative order, which is what a splice
+  // leaves behind too, so the result matches the one-at-a-time sequence.
+  const slots: Array<T | undefined> = [...items];
+  for (const addition of additions) {
+    const ref = addition.ref;
+    const known = positions.get(ref);
+    const replaced = known?.shift();
```
baseballyama: [P2] Use a head cursor rather than shifting the positions array. With N existing hyperlinks sharing one ref and N replacements of that ref, `known` stays length N because each shift is followed by a push. Every shift moves the remaining positions, so the batch is still O(N²) for the duplicate-ref inputs this helper explicitly supports. In an isolated local run, 20k entries took about 52 ms and 40k about 202 ms. The ref-read counter test cannot detect this work because it occurs inside the positions array. Store a per-ref head index and consume positions by indexed access so the duplicate path also remains linear.

.changeset/olive-moons-shake.md
```diff
@@ -0,0 +1,20 @@
+---
+'@office-kit/xlsx': patch
```
baseballyama: [P2] Mark the new public APIs as a minor release. This changeset introduces the exported `setHyperlinks` and `setComments` functions, while CLAUDE.md specifies minor for new exports. Leaving it as patch would publish these additions as 0.14.1. Please use minor here (the separate XML bug-fix changeset can remain patch); this keeps the package pre-1.0.

baseballyama (changes requested):
Re-reviewed the batch redesign. Both regressions from the previous review now pass: replacing an occupied hyperlink slot is visible immediately, and changing a comment ref preserves first-duplicate lookup/removal. The call-local index avoids the previous persistent-cache invalidation problem.

The existing 2,500 tests and all CI jobs pass. Additional reproduction tests expose chunk-dependent acceptance in the XML change (six failing cases; the same documents pass when split after the opening root). There is also a quadratic duplicate-ref path in the batch implementation, and the new public exports need a minor changeset. Details are inline.

Please also add a short usage example for the two new batch APIs to the worksheet documentation.

src/xml/iterparse.ts
baseballyama: Fixed in 4120cc0. The bounded prologue scanner now skips comments and processing instructions and stops at the root start tag within each chunk. Regression tests cover strings, bytes, single-byte streams, and every two-chunk split, including rejection of actual declarations.

src/worksheet/ref-batch.ts
baseballyama: Fixed in 09fa36d. Each ref queue now has a head cursor, so removing a position does not shift the remaining duplicates. The new regression test counts queue movement and checks the final order after more than two cycles through 5,000 duplicates. The previous direct-mutation regression cases are also retained as tests.

.changeset/olive-moons-shake.md
baseballyama: Fixed in f452a48. The batch API changeset is now minor, and README includes examples for both new APIs. The XML bug-fix changeset remains patch.

baseballyama (approved):
Thank you so much!

PR #147, opened 2026-09-17
perf(zip): bound the inflate cache, and stop double-charging re-read entries

kibertoad:
## Summary

The random-access ZIP reader cached every entry it inflated for the lifetime of the
archive, so a finished load held the entire uncompressed package in memory. That is
the cost the reader was written to avoid. The cache is now limited to small entries
and a total ceiling, which still covers everything a load actually re-reads.

Bounding the cache means an entry can be inflated more than once, which exposed a
pre-existing accounting bug: the decompression budget charged each inflate
separately, so re-reading an entry counted its payload twice against the archive
total and could reject a legitimate file.

## Motivation

There is no issue for this. It came out of a performance audit of the read and write
paths; the other items from that audit are being opened separately.

The module header claimed peak memory stays at "compressed-archive size plus per-entry
inflate scratch rather than holding every uncompressed entry resident". The unbounded
cache made that false from the first full load.

Rather than guess at a policy, I instrumented `readEntry` and ran all 38 fixture
workbooks through `loadWorkbook`. A load re-reads **nothing but `.rels` parts, and
never more than twice** (`pivot.xlsx` and `vba-test.xlsm`, via the passthrough rels
walk). No sheet, stylesheet or shared-strings part is ever read twice. The largest
fixture reaches 1033 distinct entries in one load. So the cache earns its keep only
for parts of a few KB, and everything else was being held for nothing.

Hence: cache entries up to 64 KB, hold at most 4 MB of them.

## Changes

- `src/zip/inflate-cache.ts` holds the policy: entries at or under
  `CACHE_MAX_ENTRY_BYTES` (64 KB), `CACHE_MAX_TOTAL_BYTES` (4 MB) in total, least
  recently used evicted first. A hit refreshes an entry's recency, without which the
  `.rels` parts the cache exists for are the first to go: the passthrough scan reads
  every part before the rels walk comes back for them.
- `openZip().read(path)` returns an array the caller owns, for every entry at every
  size. See "Breaking changes".
- `DecompressionBudget` charges each path the high-water mark of any single inflate of
  it, through the recorder `beginEntryInflate(budget, path)` returns. `recordInflated`
  is gone, so an inflate path cannot record bytes without opening a session first.
  The archive limit is checked even when a retry adds no charge, so a failed
  inflate cannot leave a high-water mark that lets later attempts bypass the cap.
- A streaming worksheet keeps the part bytes its row-offset index was built from.
  Sheets are far over the cache's entry ceiling, so otherwise every
  `iterRows({ minRow })` query on one handle inflates the whole sheet again.
  The retained bytes are released when the handle is collected or the workbook
  closes; a retained worksheet handle cannot read cached rows after close.
- The central-directory filename decoder is constructed once instead of per entry,
  matching how `src/io/load.ts` already handles its latin1 decoder.

The high-water-mark rule changes what `maxTotalUncompressedBytes` bounds, which should
be a deliberate decision rather than a side effect of bounding the cache. It was a
running total of inflate work; it is now a bound on distinct payload, which is what
`checkDeclaredTotals` already enforces from the central directory. A caller looping
reads over one 400 MB entry would previously trip the archive cap and now will not. The
per-entry cap and the ratio cap still apply in full to every individual read. Calling it
out because `SECURITY.md` names decompression bombs in scope and the change is not
visible in the shape of the diff.

Charging a mark rather than refunding a previous charge is also what makes overlapping
reads of one entry correct. A refund performed when a stream is constructed lands before
that stream charges anything, so two streams over one entry, which is what two
`iterRows` calls over a sheet produce, still billed it twice.

## Testing

The combined branch includes the merged #144 and #145 changes. Added regression
coverage for repeated reads of an entry with forged sizes (DEFLATE and STORE,
synchronous and streaming reads), and for indexed row reads after workbook close.
Both regressions failed before their respective fixes and pass afterward.

`tests/zip/inflate-cache.test.ts` unit-tests the cache policy directly (size ceiling,
total ceiling, eviction order, recency on a hit, copy in and copy out) and covers the
budget through `openZip`.

Three cases pin behaviour that is easy to get wrong, and each was confirmed to fail
against the implementation it replaces:

- two `readStream`s opened over one entry before either is drained, and the same two
  interleaved chunk by chunk, under an archive budget with room for one copy of it.
  Both throw `OpenXmlDecompressionBombError` if the budget refunds at stream
  construction instead of charging a mark.
- `tests/perf/row-index.test.ts` measures a repeat band query on one worksheet handle:
  14.9x faster than the first query with the part retained, 2.0x without it, gated at
  3x under `PERF_ROW_INDEX_GATE=1`.

Memory, measured on Node 24.18, Windows, loading a 10-sheet 8.3 MB workbook (300k
rows). Peak `process.memoryUsage().external`, which is where inflated `Uint8Array`
payloads live, over a 18.8 MB baseline. Three runs each, lower is better.

| | peak external | held payload |
| --- | --- | --- |
| before | 83.4 / 83.4 / 88.6 MB | about 65 MB |
| after | 39.5 / 39.5 / 39.5 MB | about 21 MB |

What remains is the compressed archive plus the sheet currently being parsed.

Worth being precise about what this does **not** move: the smallest
`--max-old-space-size` that completes the load is 416 MB either way. Inflated entries
are `Uint8Array`s in external memory, not the JS heap, and the heap is dominated by the
parsed `Worksheet` model, which this does not touch. If you are tuning `--max-old-space-size`
this changes nothing; if you are watching RSS or a container memory limit, it is worth
about 44 MB on this workload.

Reproduce:

```sh
pnpm test tests/zip/inflate-cache.test.ts
PERF_ROW_INDEX_GATE=1 pnpm test:perf tests/perf/row-index.test.ts
```

One test fix rides along, unrelated to everything above.
`tests/io/issue-28-workbook-to-buffer.test.ts` compared two saves of one workbook byte
for byte with no `mtime` on either call, so fflate stamped each entry header with the
wall clock at ZIP's two-second resolution and the comparison failed whenever the two
saves landed either side of a tick. It did exactly that on `ubuntu-latest, 22` in this
PR's matrix while the other eight jobs passed. Both saves now pin the same `mtime`,
the way `tests/io/deterministic-bytes.test.ts` already does.

## Breaking changes

`openZip().read(path)` returns a fresh array on every call. It previously handed back
the same array once an entry had been read, so mutating one read's result changed what
later reads of that path returned. Anyone relying on that aliasing, or on two reads
yielding the identical object, has to keep their own reference. The bytes are
unchanged, and the array has always been documented as safe to mutate, which is now
true of every read rather than only the first.

The changeset is `minor` and flags it.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #151, opened 2026-09-17
fix(streaming): bound shared-string retention for large exports

### LINKED ISSUE #148: feat: bound the write-only shared-strings table for high-cardinality exports
## Problem

Provenance, so it is not overstated: this came out of a performance audit of the read and
write paths, not a production incident. The numbers below are measured, the use case is
reconstructed rather than reported by a user who hit it.

`createWriteOnlyWorkbook` exists so an export can be larger than memory. It delivers that
for repeating data and not for distinct data, because every plain-string cell goes into the
shared-strings table and nothing ever leaves it.

Retained heap after writing N rows to a discard sink, read after a forced GC while the table
is fully populated, before `finalize()`. Four columns per row, two of them strings. Node
24.18, Windows.

| rows | 200 distinct strings | every string distinct |
| --- | --- | --- |
| 500k | 8 MB | 107 MB |
| 1M | 8 MB | 209 MB |
| 2M | 8 MB | 398 MB |

The repeating column is flat at 8 MB no matter how many rows go through it, which is the
behaviour the API advertises. The distinct column grows at roughly 200 MB per million rows,
about 100 bytes per unique string across `entries[]` and the `index` Map. Extrapolated, a
10M-row export of order ids or email addresses retains something like 2 GB.

This is not hypothetical shape-fitting: ids, emails, UUIDs, ISO timestamps and free-text
descriptions are the normal contents of a large export. The bigger the export, the more
likely its strings are distinct, so the failure mode gets worse exactly where the API is
supposed to help.

## Existing paths considered

- `createWriteOnlyWorkbook` from `@office-kit/xlsx/streaming` is the intended path and is
  otherwise well behaved: rows flush through a 64 KB buffer into the deflate stream, and the
  8 MB figure above shows nothing else accumulates.
- There is no way to ask for `t="inlineStr"` output. `serializeCell` routes every
  `typeof value === 'string'` through `addSharedString` unconditionally, and
  `WriteOnlyOptions` has no related knob. `grep -rn inlineStr src/` finds it only on the read
  side (`src/streaming/read-only.ts`, `src/utils/inference.ts`); nothing writes it.
- The escape hatches do technically cover it: `createZipWriter` from `@office-kit/xlsx/zip`
  plus `@office-kit/xlsx/packaging` let you assemble a package by hand. That means
  hand-writing worksheet XML, the manifest, the rels and the workbook part, which is the
  whole library. This looks like the case CLAUDE.md describes as reachable "only at an
  abstraction so low that every real user re-implements the same wrapper."

## Proposed solution

Two shapes seem defensible. I lean towards the second and would rather agree the direction
before writing it.

**A. An explicit option.**

```ts
const wb = await createWriteOnlyWorkbook(sink, { inlineStrings: true });
```

Predictable, and it keeps one canonical path (an option on the existing entry point, not a
second function). The caller decides, knowing their own cardinality. The cost is that they
have to know: the wrong choice either keeps the ceiling or inflates the file, and the docs
have to explain shared strings to someone who just wants to write rows.

**B. Bounded table with automatic overflow.**

Keep deduplicating until the table reaches a cap, then emit later strings inline. No API
surface, bounded by construction, and low-cardinality exports keep the full dedup benefit.

```ts
// no caller-visible change; inside the write-only serialiser
const SST_MAX_ENTRIES = 100_000; // or a byte budget
```

Worth flagging one interaction: #137 pinned the ZIP timestamp so identical input produces
identical bytes. A threshold keeps that property, since the same input crosses the threshold
at the same row. But two similar workbooks could land on different output shapes, and if that
is unacceptable, A is the safer answer.

## Alternatives

- **Do nothing, document the ceiling.** Honest but not actionable: there is currently no
  option for a caller who hits it, short of the escape hatches above.
- **Always emit inline strings on the write-only path.** Simplest, and defensible given this
  path is for large exports. Rejected because it is a real regression for the repeating case,
  where dedup is doing useful work and the file would grow.
- **Let the caller supply the table.** Maximum control, leaks an implementation detail into
  the public API for no benefit the options above do not already give.

## Scope and compatibility

Additive either way, no existing path changes behaviour. Under A the default stays exactly as
today. Under B output shape changes for high-cardinality exports, which is visible to anyone
diffing bytes against an older version, though not to any consumer that reads the file.

Mixing `t="s"` and `t="inlineStr"` within one sheet is legal under ECMA-376 and both Excel and
LibreOffice accept it, which B relies on. I have not yet verified that against the
conformance suite in this repo, and would do so before implementing.

## Confirmation

- [x] I have read the project's "one way to do one thing" policy and confirmed
      this proposal is not a parallel path to an existing capability.
- [x] I have searched existing issues (open and closed) and confirmed this is
      not a duplicate.
- [ ] I have a real, specific use case for this, not a hypothetical "could be
      useful." Not ticked, deliberately: this came from an audit, so the use
      case is reconstructed. The measured ceiling is real; a user hitting it is
      inferred. Weigh it accordingly.
- [x] If I used an LLM to draft this issue, I have read and verified every
      claim, including the "existing paths" section, and am willing to defend
      the proposal in follow-up.

baseballyama:
## Summary

Bound write-only string retention without adding a public option. Once a workbook-wide table reaches 100,000 entries or its 8 MiB string/XML payload budget, new values are written inline while existing shared-string IDs remain reusable across sheets.

## Motivation

Closes #148.

Distinct strings currently accumulate for the entire export. Reproducing the issue with four columns (two distinct strings), a discard sink and forced GC on Node 26.6/macOS showed heap growth from 44.8 MiB at 250k rows to 162.0 MiB at 1M rows on 0.15.0. The same workload with this change stays at 24.3–24.4 MiB from 250k to 1M rows. Final compressed output in that workload changed from 25.6 to 23.8 MiB; file-size effects depend on the data.

## Changes

- Apply the bounded policy to plain strings and rich text. The first value that does not fit permanently closes admission; registered values keep their IDs. The payload budget counts lookup keys and serialized text at two bytes per UTF-16 code unit, while the entry cap bounds container overhead. It is not a total heap limit.
- Snapshot rich-text runs and reuse the existing text/font serializer for shared and inline cells. Stream the final shared-string table in small UTF-8 chunks.
- Decode inline OOXML text escapes per text element in the streaming reader, preserving literal escape-looking text across adjacent rich-text runs.
- Document memory scope and output compatibility, correct the README example to fit Excel's per-sheet row limit, and add a minor changeset. Full policy: [docs/write-only-strings.md](docs/write-only-strings.md).

## Testing

- Added the entry-cap regression first and observed it fail on the previous implementation.
- Added byte-boundary, permanent overflow, cross-sheet ID reuse, rich-text snapshot, whitespace/control/literal-escape/Unicode, deterministic output, and UTF-8 chunk-boundary coverage. Mixed shared/inline output passes OPC, XSD and semantic validation and round-trips through both readers.
- Added a discard-sink retained-heap gate for distinct plain and rich strings. At 250k/750k rows the heap deltas were 15.1/15.0 MiB (plain) and 4.7/4.8 MiB (rich).
- Passed `npm run lint`, `npm run typecheck`, `npm run knip`, `npm test` (2,519 tests), `npm run build`, `npm run size`, `npm run check:attw`, and `npm run check:consumer`.
- Passed `PERF_GATE=1 npm run test:perf` (10 tests); the existing numeric throughput gate measured 1.46M cells/s. npm runs the same package scripts as pnpm. Desktop Excel/LibreOffice were not used for validation.

## Breaking changes

No public API or cell-value changes. Write-only ZIP bytes, shared-string IDs and raw XML representation can change, including deduplication of equivalent rich-text XML. Consumers inspecting raw XML must accept both shared and inline strings. Model-based saves are unchanged. This compatibility change is recorded as a pre-1.0 minor release; it does not require v1.

## Checklist

- [x] If this is a user-visible change, I have run `pnpm changeset` and committed the result. (Equivalent changeset file added directly.)


baseballyama:
Thank you so much!

PR #153, opened 2026-09-17
fix: emoji and other astral characters were written as _xD83D__xDE00_

kibertoad:
## Summary

Every emoji, CJK Extension B ideograph, and other astral-plane character this
library wrote reached the file as `_xD83D__xDE00_`. Excel, LibreOffice and
Google Sheets display that escape text literally. Adding the `u` flag to the
illegal-codepoint regex fixes it. The same class also left U+FFFE and U+FFFF
raw, which no conforming parser accepts, so they are escaped here too. This
edits the same function as #154; see Merge order at the bottom.

## Motivation

No issue; found while auditing the escape layer.

`escapeCellString` replaces codepoints XML 1.0 cannot carry with `_xHHHH_`. Its
character class includes the surrogate range, and without the `u` flag a regex
matches UTF-16 code units, so both halves of a well-formed pair matched
separately and each escaped on its own. `setCell(ws, 1, 1, 'hi 😀')` produced
`<t>hi _xD83D__xDE00_</t>`.

Everything on the write path was affected: plain strings and rich-text runs via
`sharedStrings.xml`, and formula bodies and cached string results via the
worksheet part.

The reason this survived is that `unescapeCellString` puts the pair back
together, so our own load-save-load round-trip returned the right string while
the file on disk was wrong. Any test that only compares the model before and
after is blind to it, which is why the new assertions read the emitted part
instead.

U+FFFE and U+FFFF sit outside XML 1.0's `Char` production, and the same class
passed them through unescaped. A cell carrying either one produced a part that
`saxes` rejects with `disallowed character`, and that `xmllint` and Excel reject
the same way. It is one token on the line this PR already rewrites, so it ships
here rather than as a second pass over the same regex.

## Changes

- Astral characters are written verbatim in cell text, rich-text runs, formula
  text and cached formula results.
- U+FFFE and U+FFFF are escaped as `_xFFFE_` / `_xFFFF_`. The astral
  noncharacters (U+1FFFE and friends) are inside `[#x10000-#x10FFFF]`, so they
  stay verbatim.
- Unpaired surrogates still escape to `_xHHHH_`. They have no UTF-8 encoding
  and are illegal in XML, so they have to.
- No reader change. A file already written with the split form reads back
  correctly wherever the reader inverts the `_xHHHH_` convention, which covers
  cell text and rich-text runs. `<f>` and `<v t="str">` are handed back as
  written, so an emoji split there stays split.

## Testing

```sh
pnpm vitest run tests/utils/escape.test.ts tests/io/astral-cell-text.test.ts \
  tests/streaming/bounded-strings.test.ts
```

Six of the new tests fail against `main`'s `escape.ts`.
`tests/io/astral-cell-text.test.ts` unzips the saved package and asserts on the
bytes of `sharedStrings.xml` and `sheet1.xml`, since asserting on the reloaded
model passes either way.

Two of them are worth a reviewer's attention:

- The streaming shared-string writer cuts its output every 16,384 UTF-16 code
  units and steps back one unit when a high surrogate lands on the cut. While
  every surrogate was being escaped to ASCII that branch could never be taken;
  this change makes it load-bearing. The new case in
  `tests/streaming/bounded-strings.test.ts` crosses the boundary at both
  parities, and fails on U+FFFD with the `end--` removed.
- The conformance case in `tests/conformance/writer-features.test.ts` puts
  astral text and all four `_xHHHH_` classes through the XSD, so the U+FFFE
  gap above cannot come back unnoticed. `xmllint` is not installed on my
  machine, so that tier is verified in CI; locally I parsed every emitted part
  with `saxes` instead, which is what produced the `disallowed character`
  diagnosis in the first place.

`tests/utils/escape.test.ts` held literal control bytes, NUL among them, which
is what made Git classify it as binary and render every change to it as
`Bin 2085 -> 3026 bytes`. They are `\uXXXX` escapes now, so the file is plain
text from here on, and `.gitattributes` marks `*.ts` as diffable in case a
control byte lands in one again.

It still renders as binary on the Files tab, though: the NULs sit on the `main`
side of the comparison, and a `.gitattributes` on the head does not change how
the base blob is classified. Read it via "View file" or `git show
fix/astral-chars-in-cell-strings:tests/utils/escape.test.ts`. The file holds 24
assertions over 12 cases, 10 assertions and 4 cases of that new here: astral
characters through escape and unescape, an unpaired surrogate still escaping,
U+FFFE / U+FFFF escaping while U+1FFFE does not, and the split form still
unescaping for files already on disk.

## Breaking changes

`escapeCellString` is exported from `@office-kit/xlsx/utils`, and its output
changes for astral characters, U+FFFE and U+FFFF. The changeset is `minor`,
which is what CLAUDE.md asks for on a pre-1.0 behavior change to a documented
export. Readers are unaffected. A consumer diffing writer output against a
stored file that contains `_xD83D_` will see it change.

## Checklist

      the result.
      marked accordingly.
      comments.

## Merge order

#154 drops `escapeCellString` from `<f>` and `<v t="str">` altogether, which is
the deeper fix for the formula paths: nothing inverts `_xHHHH_` there, so the
convention should never have been applied to them. Both branches edit
`src/utils/escape.ts` and will conflict. This one is the smaller change and
expects to land first.

For whoever merges #154: its `escapeXmlTextVerbatim` rejects the C0 controls
and unpaired surrogates, but not U+FFFE / U+FFFF, so a cached formula result
carrying either one would reach the file raw again. The class in this PR closes
that only for the shared-string path.


baseballyama (approved):
Thank you so much!

PR #154, opened 2026-09-17
fix: formula text containing a literal _xNNNN_ was rewritten on every save

kibertoad:
## Summary

Formula text and cached formula results were escaped with the shared-string
escaper, which rewrites a literal `_xNNNN_` into `_x005F_xNNNN_` on every save
even though nothing reverses it on read. `CONCAT("_x0041_")` was wrong in the
first file written and grew another prefix on each cycle after that. Those
nodes now go through one escaper built for text the reader hands back verbatim:
XML escaping, a character reference for CR, and a hard error for a codepoint
XML 1.0 cannot carry. Two cached-result typing bugs on the same path go with
it.

## Motivation

No issue; found while auditing the escape layer.

`escapeCellString` bundles two independent jobs: protecting an already-escaped
`_xHHHH_` sequence by turning its leading underscore into `_x005F_`, and
encoding the codepoints XML 1.0 cannot represent. Only the first belongs to
shared-string text, where the reader reverses it. `<f>` gets no reversal at
all, so applying the protection pass there was one-way rewriting. Dropping the
whole escaper rather than that one pass would trade the bug for a worse one:
a control character then reaches the file raw and the package is no longer
well-formed XML, which the library's own streaming reader rejects.

The same two nodes carry a formula's cached result, and the reader and writer
disagreed about its `t` attribute: a result read from `t="s"` or `t="e"` was
written back as `t="str"`.

## Changes

- Formula text (`<f>`) and a `t="str"` cached result keep a literal `_xNNNN_`
  verbatim, across any number of load-and-save cycles.
- `<formula>` (conditional formatting) and `<formula1>` / `<formula2>` (data
  validation) use the same escaper; they carried the illegal-codepoint hole
  already.
- A carriage return is written as `&#13;` and reads back as a CR. A raw CR is
  normalised to LF by any conforming parser, including both of ours.
- A codepoint XML 1.0 cannot represent (a C0 control other than tab / LF / CR,
  an unpaired surrogate, or U+FFFE / U+FFFF) throws `OpenXmlSchemaError` naming the cell,
  instead of being encoded as an `_xHHHH_` sequence nothing reverses.
- A cached error result keeps `t="e"`, so `ISERROR` / `IFERROR` and
  error-keyed conditional formats still match it before Excel recalculates.
  The optional `cachedValueType: 'error'` preserves that distinction on all
  formula kinds; an ordinary cached string such as `"#N/A"` remains text.
- A cached result held in the shared-strings table (`t="s"`, which non-Excel
  producers write) resolves to its text. It used to load as the raw sst index
  and save back as that number.

## Testing

```sh
pnpm vitest run tests/formula/
```

- `formula-underscore-escape.test.ts` asserts on the emitted `sheet1.xml` and
  drives three load-and-save cycles, checking the formula text and the cached
  result on each one.
- `formula-text-codepoints.test.ts` pins the CR / LF / tab and astral
  round-trips, and the three rejection paths.
- `formula-cached-value-types.test.ts` pins the `t="e"` and `t="s"` mappings.
  Its `t="s"` fixture is assembled by hand: nothing in the public API writes a
  formula whose cached result lives in the sst.

Additional maintainer regression tests reproduce error-looking strings being
misclassified and U+FFFE / U+FFFF escaping validation. All formula kinds retain
the explicit error marker after round trips.

`validateXlsx`'s XSD tier is skipped on my dev box (no `xmllint` available on
Windows), so the schema check on the new `t="e"` shape rides on CI.

Performance gates now run one file at a time so throughput and heap tests do
not compete for CPU. All gates and their thresholds are unchanged.

## Breaking changes

Formula options gain the optional `cachedValueType: 'error'` marker without
changing the existing `cachedValue` types. The escaping behaviour changes, so the changeset is a
`minor` with the behaviour change called out at the top.

Saving now throws where it previously wrote a corrupt file: a formula or cached
string result carrying a C0 control character or an unpaired surrogate. Such a
value can only come from a caller, never from a loaded file, since neither
shape survives a parse.

Formulas in files written by earlier versions keep whatever `_x005F_` prefixes
they have already accumulated. Healing them on read is not possible: a stored
`_x005F_x0041_` is indistinguishable from one Excel wrote deliberately, so
unescaping would corrupt legitimate input. They stop growing, which is the part
that can be fixed here.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #155, opened 2026-09-17
fix(zip): reject entries whose declared bytes are not in the archive

kibertoad:
## Summary

Two central-directory shapes let an archive describe entry bytes it does not
contain, and the reader trusted both. A DEFLATE entry declaring zero compressed
bytes sent `readStream` into a busy loop that could not be cancelled. An entry
whose declared size runs past the end of the file was read from whatever bytes
happened to be left. Both now fail with `OpenXmlIoError`.

## Motivation

No issue; found while auditing the zip reader against the threat model
`SECURITY.md` states.

`readEntryStream` drives fflate's `Inflate` from `pull()`. With no bytes to
push, the loop body never ran, the inflater never saw a final chunk, and `pull`
returned having enqueued nothing, closed nothing and errored nothing. The
stream machinery then called `pull` again with identical state. There is no
timeout and no cancellation path out of it: the read never settles.
`loadWorkbookStream` reads worksheet parts through that stream, so a crafted
archive reaches the loop.

The second shape is quieter. `readCompressedBytes` ends in
`subarray(dataStart, dataStart + compSize)`, and `subarray` clamps rather than
throwing, so an entry declaring more bytes than the archive holds yields a
short slice instead of an error. XML parts fail downstream on a parse error,
but binary passthrough parts have no parser: a truncated `vbaProject.bin` or
image was captured and written straight back out on save.

## Changes

- `read` and `readStream` reject an entry whose declared compressed span ends
  past the end of the archive.
- `read` and `readStream` reject a method-8 entry carrying zero compressed
  bytes. RFC 1951 gives every deflate stream at least one block, and the last
  one carries BFINAL, so there is no stream to decode.
- Both paths reject a deflate stream that drains without ever reporting its
  final block.
- `readStream` throws synchronously, matching how it already reports an unknown
  path.
- Removed the `gpFlag` field on `CdEntry`, parsed and stored for every entry
  and read nowhere.

## Testing

```sh
pnpm vitest run tests/zip/
```

`tests/zip/missing-entry-bytes.test.ts` drives the public `openZip` and
`loadWorkbookStream`. Five of its six cases fail against `main`, and the
streaming-workbook case times out there rather than asserting, which is the
hang itself. The sixth case is the legitimate neighbour a guard must not
break: a zero-byte file does get a deflate stream, one that emits no bytes and
has to close cleanly on both read paths.

Fixtures come from the project's own `createZipWriter` with a single header
field rewritten afterwards, through a new `tests/zip/_helpers.ts`.
`tests/zip/decompression-guard.test.ts` moves onto those helpers in place of
its own copy, so the zip record layout is encoded once.

## Breaking changes

An archive carrying either entry shape used to load and now throws. A scan of
the archives in this repository (39 files, 1587 entries) found no instance of
either, and I know of no writer that emits one, so I expect no real file to be
affected. The blast radius if one is: `capturePassthrough` reads every part
unconditionally, so a single bad entry fails the whole load even in a part
nothing needs, such as `xl/printerSettings/printerSettings1.bin` or
`customXml/*`. There is no flag to skip it; the archive has to be repaired or
re-saved. The changeset is `patch` and states the rejection.

## Notes for the reviewer

The risk is that blast radius, not the guards themselves. Rejecting is the
behaviour `CLAUDE.md` asks for on invalid external input, and the alternative,
returning an empty or short part, is what let the corrupt bytes through in the
first place. A per-part skip option would be a wider API question than this
fix.

A truncated deflate *stream* (bytes present, stream incomplete) was never the
spin vector: fflate raises "unexpected EOF" on the final push and the existing
`catch` in the pull loop wraps it. I checked that directly against fflate's
`Inflate` at several truncation points. The new completeness check sits behind
that as structure rather than as a second diagnosis, so `pull` has three
exhaustive outcomes (enqueue, close, error) instead of leaving a fourth
unreachable by caller-side precondition. That is also why the close condition
is now one `settleWhenDrained` closure rather than two copies.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #156, opened 2026-09-17
fix: a workbook-level part was read from the wrong path, or not at all

kibertoad:
## Summary

`loadWorkbookStream` located `sharedStrings` and `styles` by hardcoded path, and
`loadWorkbook` preferred that path over the workbook relationship. Both loaders
now follow the relationship, which is what binds a part to the workbook, and a
relationship that resolves to nothing fails loudly instead of reading as an
absent part.

## Motivation

No issue; found while comparing the two loaders.

`xl/sharedStrings.xml` and `xl/styles.xml` are a convention. What binds a part
to the workbook is its relationship, and a producer is free to name the part
something else. The streaming loader never consulted the rels, so a workbook
that stores either part elsewhere streamed back `null` for every shared-string
cell and resolved every style against an empty pool. `loadWorkbook` did consult
them, but only after the conventional path, so a part left at that path shadowed
the one the rels name.

The failure mode is the bad one in both cases: no error, no warning, just the
wrong data. A caller has no way to tell that apart from a genuinely empty sheet.

The streaming loader's sheet list had the same shape of silence. A sheet whose
`r:id` had no rels entry was dropped from `sheetNames`, a duplicate sheet name
kept the last part and made the first unreachable, and a workbook that declares
sheets with no rels part opened with no sheets at all. `loadWorkbook` rejects
all three.

## Changes

- Both loaders resolve `sharedStrings` and `styles`, and `loadWorkbook` also
  resolves the theme, through the workbook relationships, falling back to the conventional path only when there
  is no such relationship.
- A workbook-level relationship that targets a part the package does not
  contain, or that carries `TargetMode="External"`, throws an
  `OpenXmlSchemaError`.
- Optional workbook parts prefer the exact ZIP entry name. If it is absent,
  percent-decoding supports `Target="shared%20strings.xml"` naming
  `xl/shared strings.xml`, while preserving existing encoded entry names.
- `loadWorkbookStream` parses `<sheets>` with `loadWorkbook`'s parser and
  rejects the same declarations: an `r:id` with no rels entry, a duplicate sheet
  name, and a sheet list with no rels part.
- One internal `readOptionalWorkbookPart` serves all three parts and takes a
  part descriptor, so a call site cannot swap two adjacent `string` arguments
  and silently get `undefined`.

## Testing

```sh
pnpm vitest run tests/io/non-canonical-part-paths.test.ts tests/streaming/non-canonical-part-paths.test.ts
```

Both fixtures start from `workbookToBytes` output and move the parts within it,
so they carry what the writer emits and follow it when that changes.
`tests/io/` covers the lookup; `tests/streaming/` covers the parity claim,
asserting that both loaders reject the same malformed package for the same
stated reason.

## Breaking changes

Packages that used to load now throw, in the cases where what they loaded was
wrong:

- a workbook-level relationship pointing at a part the package does not contain,
  or an external one. The part previously read as absent, so a whole
  shared-string table or style pool went missing without a word.
- the three malformed sheet declarations above, where `loadWorkbookStream`
  previously returned a shorter or wrong sheet list.

No well-formed package is affected, and each new rejection matches what
`loadWorkbook` already did for equivalent input. The changeset stays a patch:
pre-1.0, and every path that starts throwing produces silently wrong reads
today.

## Notes for the reviewer

`readOptionalWorkbookPart`, `parseSheetEntries` and `OFFICE_DOC_REL_TYPE` are
exported from `src/io/load.ts` but deliberately not re-exported from
`src/io/index.ts`, so they stay internal rather than becoming public API. That
follows `resolveRelTarget` and `parseDate1904`, which the streaming loader
already imports the same way.

Percent-decoding is limited to a fallback in `readOptionalWorkbookPart`.
Unconditional decoding in the global resolver broke packages with encoded ZIP
entry names. Regression tests cover both spellings and exact-match precedence;
other relationship resolution keeps its existing behavior.

`tests/io/non-canonical-package.ts` is a fixture builder imported from
`tests/streaming/`, following `tests/conformance/validate.ts`.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #157, opened 2026-09-17
fix: boolean cells written as true/false were read as FALSE

kibertoad:
## Summary

`<c t="b"><v>true</v></c>` read back as FALSE, and so did a padded
`<v>\n  1\n</v>`. `t="b"` carries an xsd:boolean, whose lexical space is `1` /
`0` / `true` / `false` and whose whiteSpace facet is `collapse`, but both
loaders compared the text against `"1"`. Every form now reads correctly, in
cell values and in cached formula results alike.

## Motivation

No issue; found while auditing the cell-value readers.

Excel writes `1` / `0`, which is why this went unnoticed, but the long spelling
is equally valid and other producers emit it. The reader's own attribute parser
already accepted all four forms, as do the data-table flags and every chart
parser in the codebase, so the cell-value path was the outlier. Nothing was
raised on a mismatch; the value was just wrong, which makes it the hardest kind
of bug for a user to notice.

The whitespace half matters because `src/xml/parser.ts` runs fast-xml-parser
with `trimValues: false`. A producer that indents its parts, or a file that has
been through `xmllint --format`, hands the reader `<v>\n  1\n</v>` with the
padding intact.

## Changes

- Every lexical form reads correctly, with or without insignificant
  whitespace, in cell values and cached formula results, in `loadWorkbook` and
  `loadWorkbookStream` alike.
- One `parseXsdBoolean` in `src/utils/xsd-boolean.ts` backs both cell readers,
  the worksheet attribute parser and the schema attribute coercion. It stays
  internal: no subpath re-exports it.
- A `t="b"` cell with no value reads as empty rather than FALSE, matching how
  `t="n"` treats a missing value, and saves as an untyped `<c/>`.
- A value outside the lexical space makes `loadWorkbook` throw
  `OpenXmlSchemaError` naming the cell; `loadWorkbookStream` reads it as empty,
  staying lenient the way it is for every other unreadable value. The README
  now states that split, which was undocumented.

### Decisions a reviewer may want to push on

- **Mixed case and `t` / `f` are accepted, though they sit outside the lexical
  space proper.** `src/schema/serialize.ts` accepts them for openpyxl parity
  and `cellValueAsBoolean` lowercases before comparing. Rejecting them in the
  cell reader alone would refuse a file that openpyxl and Excel both open, and
  rejection is fatal here, so the wider set wins.
- **A junk cached result throws instead of being dropped.** Dropping it loses
  the cached value and the `t="b"` the writer derives from it, so a load then
  save would rewrite the cell as an untyped formula.
- **`parseBoolXmlAttr` is renamed rather than wrapped.** It reads `<v>` text as
  well as attributes, so the old name was wrong. That rename is the ~30 lines
  of call-site churn in `reader.ts`.
- **No `strict` flag on `LoadOptions`.** The two loaders' contracts are
  documented instead; a switch would be a second path through the same
  capability, and nobody has asked for one.

Deliberately left out: the roughly thirty other hand-rolled
`raw === '1' || raw === 'true'` sites in the chart, table and stylesheet
readers. They share the whitespace blind spot and belong on `parseXsdBoolean`,
but that is a sweep of its own and none of it sits in this bug's path.

## Testing

```sh
pnpm vitest run tests/worksheet/boolean-cell-lexical-forms.test.ts
```

The padded and upper-case inputs run through `loadWorkbook` and
`loadWorkbookStream` end to end rather than through `parseWorksheetXml` alone,
since `trimValues: false` is the reason they read differently. `<c t="b"/>` is
asserted through a save, so the dropped type attribute is pinned instead of
incidental.

## Breaking changes

No API change. Three input shapes are read differently, and all three were
wrong before:

- `<v>true</v>` / `<v>false</v>` give TRUE / FALSE instead of both FALSE.
- `<c t="b"/>` gives `null` instead of `false`, and saves as `<c/>`: with no
  value there is no boolean for `t="b"` to describe.
- `<c t="b"><v>yes</v></c>` throws from `loadWorkbook` instead of giving
  `false`; `loadWorkbookStream` gives `null`.

Boolean attributes across the worksheet, chart and style parts also accept
padded, mixed-case and `t` / `f` spellings now, which only turns
previously-ignored attributes into honoured ones.

The changeset is `minor`, per CLAUDE.md's rule for a pre-1.0 behavior change.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #158, opened 2026-09-17
fix: worksheets whose rows omit @r read as empty, or not at all

kibertoad:
## Summary

`@r` is optional on `<row>`, and a worksheet that omits it broke both loaders in
different ways: `loadWorkbook` rejected the file, and `loadWorkbookStream`
reported an empty sheet with no error. Such a row is now placed by its first
located cell, or by the row after the highest one read so far, in both loaders,
and a streaming band query covers it.

## Motivation

No issue; found while auditing the row-iteration paths.

ECMA-376 section 18.3.1.73 marks `@r` optional on CT_Row. Excel always writes
it, so the two readers had settled into two different wrong answers for
generators that do not:

- `loadWorkbook` threw `worksheet: <row> missing required @r`.
- `loadWorkbookStream` derived the number from `currentRow`, which is back at
  its `-1` sentinel between rows. Every affected row therefore landed on row 0,
  fell below the default `minRow` of 1, and was dropped. The iterator yielded
  nothing and raised nothing.

The second is the worse of the two: the caller sees a valid, empty-looking
worksheet.

Three places number a row: the DOM parser, the streaming SAX walk, and the
byte-level row-offset index a band query seeks with. The index reads `<row>`
tags and nothing else, so a rule that depends on a row's cells has to leave the
index unable to seek. Otherwise it jumps to a row it numbered differently from
the walk, and the rows in between vanish with no error, which is the same
failure this PR exists to remove.

## Changes

- A `<row>` without `@r` takes the row named by its first cell that carries one.
  Cell refs are the only record such a generator leaves of a gap, and openpyxl's
  reader places the row the same way. With no located cell, the row takes the
  slot after the highest row read so far, so an out-of-order `@r` cannot send a
  later bare row onto a row already read and overwrite it.
- `loadWorkbook` accepts an `@r`-less row whose cells record a gap, instead of
  rejecting it with an error quoting an `@r` the file never carried.
- Streaming band queries (`minRow` / `maxRow`) cover those rows. A sheet holding
  a row the byte index cannot number is treated as unseekable: its band queries
  stream the part and retain nothing, rather than seeking by a number the walk
  disagrees with.
- `<row r>` outside `[1, 1048576]` throws from both readers, as does a derived
  row that would fall past the last one. A malformed `@r` such as `r="nope"`
  throws in the streaming reader too, instead of silently shifting every row
  after it.
- `readRowAttr` (bytes) and the string parse accept the same shapes, including
  an optional plus sign and the whitespace `xsd:unsignedInt` collapses, so the index and the walk cannot
  read one `@r` two ways.

Unlocated cells preceding the first located cell are buffered for that row
until its number is known, then filtered. Column numbering is independent of
column filters. Regression tests cover both cases and signed row attributes.

## Testing

```sh
pnpm vitest run tests/worksheet/row-without-r-attribute.test.ts
```

Two cases are worth naming because they pin the invariants a reader would
otherwise have to take on trust: every band query is compared against the full
walk of the same sheet, over each shape of mixed and bare sheet, and the two
loaders are compared cell by cell on those same sheets.

Local runs fail the `tests/conformance/` XSD tier with `xmllint not available`
on a machine without `libxml2-utils`; CI installs it.

## Breaking changes

No API change, and no conformant file reads differently. Two malformed inputs
that used to load now throw an `OpenXmlSchemaError`: a `<row r>` past row
1048576 on a row with no cells (`setCell`'s range guard never saw it), and a
`<row r>` that is not a number at all in the streaming reader, which used to
drop the row and renumber everything after it. Neither is a row any reader can
place, so the changeset stays a patch.

## Notes for the reviewer

Review effort belongs on the streaming band-query path in
`src/streaming/read-only.ts`: the numbering rule in `iterSheetRows`, the
`hasUnnumberedRow` flag out of `buildRowOffsetIndex`, and the `SheetSeek` split
that decides between seeking and streaming. The rest follows from those.

The trade-off there is deliberate. A sheet that omits any `@r` loses the seek and
pays a full SAX walk per band query. Keeping the seek would mean teaching the
byte scan to parse cell refs, and every shape those two parsers disagree on is
another way for a band query to drop rows silently. The seek is unchanged for
every sheet Excel, LibreOffice or this library writes.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #159, opened 2026-09-17
fix: a corrupt numeric cell failed at save instead of at load

kibertoad:
## Summary

A numeric cell whose `<v>` is not a finite number is refused at load, in a
message that names the sheet, the cell and the offending text.
`<c t="n"><v>oops</v></c>` used to parse to `NaN` and travel on as a cell value,
with the failure surfacing much later from the writer as "cannot serialise
non-finite number", naming a cell the caller never wrote. The writer is held to
the same rule on the one path that was missing it.

## Motivation

No issue; found while auditing the cell-value readers.

`<v>` under `t="n"` went straight through `Number.parseFloat` with no check on
the result. Text became `NaN`, an exponent past the double range became
`Infinity`, and both were stored on the cell. The load reported success, so the
first sign of trouble came on save, at the far end of the pipeline, in a message
that points at a cell rather than at the file that produced it. Diagnosing that
means working backwards from a save to a load that happened minutes earlier.

This is what CLAUDE.md's "external input was invalid, throw an `OpenXmlError`
subclass" is for: the boundary is the reader.

## Changes

- `loadWorkbook` throws `OpenXmlSchemaError` for a `t="n"` value that is not a
  finite number, both for cell values and for cached formula results. The
  message locates the cell (`worksheet: <v>oops</v> at Data!B2 is not a finite
  number`) and quotes at most 40 characters of the text, since cell text is
  attacker-controlled and unbounded.
- `loadWorkbookStream` throws the same error on the same bytes. Reading the cell
  as empty instead would be indistinguishable from an empty cell, and would leave
  the two entry points disagreeing about one file.
- `saveWorkbook` refuses a non-finite cached formula value, which it used to
  write out as `<v>NaN</v>`. Excel numeric cells cannot represent that non-finite result. The guard now sits in the one function every numeric `<v>` passes
  through, which also covers date and duration serials.
- An absent, empty or whitespace-only `<v>` is an empty cell. A pretty-printed
  part reaches the reader with its indentation inside `<v>`, so the blank test
  has to ignore whitespace. Ordinary numbers, including any surrounding
  whitespace, read as before.

## Testing

```sh
pnpm vitest run tests/worksheet/non-numeric-numeric-cell.test.ts tests/streaming/non-numeric-numeric-cell.test.ts
```

12 of the 14 cases fail against `main`. The two that pass are the ordinary
numbers that must keep working and the writer's existing guard on a plain cell
value. The streaming file covers both routes into the row iterator, since a
band query with `minRow > 1` replays from the row-offset index rather than
streaming the part.

The maintainer review also covers malformed numeric values in cells whose row
number is derived from a later cell reference. Error messages use the resolved
coordinate, and rows outside the requested band stay skipped. The integrated
changes pass type checking, lint, the full suite, package checks, and all
performance gates.

## Breaking changes

A file that used to load, carrying `NaN` or `Infinity` on a cell, now throws
from `loadWorkbook` and from `loadWorkbookStream`. That file was already
unsaveable, so the change moves an existing failure to the boundary that
produced it and gives it a message pointing at the cause. The changeset is a
`minor` for that reason.

## Notes for the reviewer

- Trailing-garbage forms that `parseFloat` accepts (`12abc` parses as `12`) are
  left alone. That is a question about the xsd:double lexical space, separate
  from the one here, which is a value the rest of the pipeline cannot represent.
  Rejecting it could turn away files that load fine today, and belongs in its
  own change if it is wanted.
- The model setters still accept a non-finite number: `setCell(ws, 1, 1, NaN)`
  stores it and fails at save with the writer's message. `setCellValue` is
  documented as "no validation beyond the union, the caller is in charge", so
  tightening the write API is a separate decision rather than part of a reader
  fix.
- An unhandled `t` in the streaming reader (`t="d"`, or a value outside
  ST_CellType) keeps its existing leniency. The type says nothing about what the
  text holds, so there is no finiteness rule to apply, and `loadWorkbook` rejects
  the cell type outright. Aligning those two is its own change.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

baseballyama (approved):
Thank you so much!

PR #161, opened 2026-09-17
test(perf): warm the throughput gate up before it measures

kibertoad:
## Summary

#154 serialised the perf suite and fixed the flaky `perf` job. This picks up the
two things it left behind: the throughput gate's first pass is still a warmup
that gets reported as a measurement, and `fileParallelism` needs a note saying
why `singleFork` is not the same knob.

## Motivation

No issue; opened after the `perf` job failed on four branches on 2026-09-17
that touch neither the writer nor the SAX walk
(`fix/optional-row-r-attribute`, `fix/streaming-loader-part-resolution`,
`fix/formula-text-escape-roundtrip`, `fix/astral-chars-in-cell-strings`).

The cause was that the suite competed with itself. Each of the six files under
`tests/perf/` is a deliberate CPU and memory stress, and vitest resolves
`availableParallelism() - 1` workers, so three ran at once on a 4-vCPU
`ubuntu-latest` and the throughput gate measured whatever share of the box it
was given. `fileParallelism: false` in a49df8e (#154) fixed exactly that, so
most of what I set out to do was already done. The three CI runs since it landed
confirm it worked:

| | before (n=25) | after (n=3) |
| --- | --- | --- |
| throughput, best of 3 | 471k to 1,091k, median 660k | 668k, 676k, 803k |
| under the 500k floor | 4 runs (16%) | none |
| `iterparse-heap` peak delta | 16 MB in 18 runs, 67 to 87 MB in 7 | 16 MB, 16 MB, 16 MB |
| `perf` job duration | 20 to 40s | 75s, 75s, 76s |

That `iterparse-heap` row is the interesting one: its ceiling is 150 MB
(`tests/perf/iterparse-heap.test.ts:47-48`), so at 87 MB it was the next gate
due to tip, and it was not a heap regression at all. A contended V8 defers
collection, so the delta captured garbage a quiet run reclaims.

What is left over:

- **The first pass is a warmup being reported as a measurement.** On those same
  three runs iteration 1 read 553k, 561k and 631k against 663k, 676k and 803k
  for the passes after it. Best-of-3 keeps the headline number honest, but the
  `runs:` detail line is what a human reads when judging whether a change
  regressed anything, and it currently shows a warmup curve rather than three
  comparable numbers. `pnpm bench` already passes `warmupIterations: 1` for this
  same workload (`tests/perf/throughput.bench.ts:37`) and vitest serialises
  benchmark mode, so the bench and the gate were measuring the same code with
  opposite methodology.
- **`singleFork` is a rake next to `fileParallelism`.** It reads like the same
  knob and would look like a way to cut the job's new 75s, but it shares one
  process across files, so a heap gate's baseline picks up whatever the previous
  file left behind, which is what `pool: 'forks'` (45d6ad1) and
  `execArgv: ['--expose-gc']` (ec45506) are there to prevent.

## Changes

- `tests/perf/throughput.test.ts`: one discarded warmup pass before the three
  measured iterations. Costs about 5s per CI run.
- `vitest.perf.config.ts`: three comment lines naming the `singleFork` rake. No
  behavior change.

Deliberately not here:

- **The 500,000 floor is unchanged.** It should go up, not down. Serialised, CI
  lands 34% above it with little spread, and a quiet 24-core box does about
  1.25M cells/s, so a 2x regression would still pass today. The right number is
  a fraction of the serialised median, and three runs is not enough to read that
  off. Follow-up once there are ten or so.
- **The heap ceilings are unchanged**, for the same reason: `iterparse-heap` has
  gone from a 150 MB ceiling against an 87 MB worst case to a flat 16 MB, so
  there is room to tighten it, but on three runs the new spread is a guess.

## Testing

```sh
PERF_GATE=1 pnpm test:perf
```

All six files pass. With the warmup pass the three measured iterations agree
within 6% on a quiet box (1,184,986 / 1,195,654 / 1,256,201 cells/s) instead of
climbing out of a cold start, which is the whole point of the change.

The default `pnpm test` suite is untouched: it uses `vitest.config.ts`, which
excludes `tests/perf/**`.

## Breaking changes

None. Test tooling only, no `src/` change, so no changeset.

## Notes for the reviewer

Worth knowing, though not addressed here: after serialising, a host-speed band
remains. Within the pre-#154 sample, vitest's total `tests` time ranged 58.59s
to 110.51s for identical work, and all four failures came from hosts at or above
96.75s. If that band eats the margin once the floor goes up, the next lever is
the runner (a larger `runs-on`, or pinning the gate to `macos-latest`, the
M-series the thresholds were originally calibrated against), not a looser
assertion.

Two gates in `tests/perf/` never run in CI, because the job sets only
`PERF_GATE=1`: `heap.test.ts:29` behind `PERF_HEAP_GATE`, and
`row-index.test.ts:23,27` behind `PERF_ROW_INDEX_GATE`. Out of scope here, but
worth knowing they are dormant. The row-index pair is ratio-based and would pass
comfortably; the heap one has less headroom than the CI numbers suggest, since
CI reads about 170,000 cells/heap-MB where a quiet local box reads 90,304.

## Checklist

- [ ] If this is a user-visible change, I have run `pnpm changeset` and committed
      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you so much!

PR #162, opened 2026-09-17
feat(worksheet)!: add value extents and rename the materialized cell extent

kibertoad:
## Summary

`getMaxRow`, `getMaxCol` and the used-range extent count any cell object in
`ws.rows`, including one that exists only to carry a style, and `iterRows` /
`iterValues` are bounded by that box. A sheet formatted down to row 200 with
values in 4 rows yields 196 rows of `null`. This adds `getValueExtent`, the
bounding box of the cells that hold a value, whose return shape is an
`IterRowsOptions`, so `iterValues(ws, box)` is the bounded walk. The used-range
helper is renamed `getCellExtent` to keep the two readings apart by name.

## Motivation

Both readings of "where does the data end" are legitimate, and Excel keeps them
apart: its used range and the `<dimension>` element it writes include
formatting, while what the author of a sheet means by "my data" is the cells
holding values. The library only offered the first, under names that read like
the second. openpyxl's inflated `max_row` is the well-known version of the same
trap.

The concrete failure is in a service that ingests user-uploaded spreadsheets:
map `iterValues` over a sheet a human has been maintaining, and the
formatted-but-empty rows arrive as phantom records. Reproduced on 32e67d0 by
saving a four-row sheet with a style and no value on rows 5 and 6, then loading
it back: `getMaxRow` returns 6 and `iterValues` yields 6 rows, the last two
entirely `null`. That round trip is a case in the new test file.

No issue: this came out of using the library rather than from the tracker.

## Changes

- `getValueExtent(ws)` in `@office-kit/xlsx/worksheet` returns the bounding box
  of the cells whose `value` is neither `null` nor `''`, or `undefined` when no
  cell on the sheet holds one. The shape is an `IterRowsOptions`, so passing it
  bounds the iteration by the whole box, min corner included:
  `const box = getValueExtent(ws); if (box) iterValues(ws, box)`.
- **Breaking:** `getDataExtent` is renamed `getCellExtent`, and
  `getDataExtentRef` (newly exported here) is `getCellExtentRef`. Behaviour is
  unchanged; see the section below.
- `getCellExtentRef` is exported from `@office-kit/xlsx/worksheet`.
  `makeAutoFilter` and `makeTableDefinition` both want an unqualified A1 `ref`
  string, `rangeToString` is not public, and `getRangeAddress` returns the
  sheet-qualified form, so without this a caller hand-rolls column letters.
- `getSheetByIndex` is exported from `@office-kit/xlsx/workbook`. It was defined
  but reachable from no subpath, so "the first sheet of an uploaded file" went
  through `sheetNames(wb)[0]` plus `getSheet(...)`, which costs two non-null
  assertions under `noUncheckedIndexedAccess`. The README's first example uses
  it now.
- `hasSheet`, `hasWorksheet` and `hasChartsheet` are deleted. Each restated in
  one expression a path the public API already has (`sheetNames(wb).includes(t)`,
  `getSheet(wb, t) !== undefined`, `getChartsheet(wb, t) !== undefined`), the
  category 8ce7059 trimmed as "one-line predicate shims". `getSheetIndex`,
  `countSheets` and `getSheetTitles` go with them: also exported from no
  subpath, and called by nothing but their own tests once `hasSheet` is gone.
  Nothing published changes.
- `sheetNames`' docstring said "All worksheet titles" while the body has always
  returned every tab, chartsheets included. Corrected, because the `hasSheet`
  deletion rests on what it actually returns.
- Docs for both extents: the README section, the cheatsheet, `llms.txt`, and the
  openpyxl migration guide, which is where the inflated `max_row` needs the
  pointer most.

Four things deliberately absent.

There is no `extent: 'cells' | 'values'` option on `IterRowsOptions`. The value
extent's box already is the options object, so an option would be a second
spelling of one capability, which "one way to do one thing" rejects. It is also
the weaker spelling: an option has to pick a default for the min corner, and
leaving it at 1 pads the head of the iteration with the same all-`null` rows
this PR removes from the tail. `@office-kit/xlsx/streaming` exports its own
`IterRowsOptions` and cannot resolve a value extent without reading the sheet it
is streaming, so the option would exist on one subpath and not the other.

A cell holding `''` counts as blank for the value extent. A trailing band of
empty shared strings is what CSV converters and non-Excel producers emit, and it
produces exactly the phantom records the extent exists to cut. The cost is that
a deliberate `setCell(ws, r, c, '')` sits outside the box; `getCellExtent`
reports it.

There is no option to drop blank rows from the middle of the yielded sequence:
that renumbers rows silently, and `filter((row) => row.some((v) => v !== null))`
is explicit and already in the `iterRows` docstring.

There is no `getValueExtentRef` to match `getCellExtentRef`. The ref string
earns its place by being the `<dimension>` form the autofilter and table
builders consume, and nothing consumes a value-extent ref.

## Testing

`pnpm test tests/worksheet/value-extent.test.ts tests/worksheet/cell-extent.test.ts tests/workbook/sheet-lookup-by-kind.test.ts tests/smoke/readme-examples.test.ts`

Three pins are worth naming, because a reviewer would otherwise have to take
them on trust:

- The round trip: a workbook whose last two rows carry a registered style and no
  value is saved, reloaded, and asserted to report `getMaxRow() === 6` with a
  value extent of 4 rows. Formatting-only cells have to survive the writer and
  the reader for the feature to matter on a real upload, and a reloaded
  empty-string cell is pinned the same way.
- `tests/workbook/sheet-lookup-by-kind.test.ts` covers the
  worksheet-versus-chartsheet discrimination in `getSheet` / `getChartsheet` /
  `getSheetByIndex` that the deleted `has*` tests were the only cover for.
- `tests/smoke/readme-examples.test.ts` runs the README's two extent calls and
  its `getSheetByIndex` example, so a later rename breaks CI instead of the
  README.

Locally the ten test files that call `validateXlsx` fail on this machine for
lack of `xmllint`, identically on `main`; CI installs `libxml2-utils`.

## Breaking changes

`getDataExtent` and `getDataExtentRef` are gone; call `getCellExtent` and
`getCellExtentRef`. Same arguments, same return shape, same behaviour, so the
fix is the import and the call site. `getDataExtent` was the only one of the two
published, and the changeset spells the rename out.

The rename is the point: with `getValueExtent` next to it, "data extent" reads
like the box that holds the data when it is the box that holds every
materialised cell. Pre-1.0 CLAUDE.md says to rename a misleading name outright
rather than parallelise around it.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you! Separating the value extent from the materialized cell extent makes formatted-but-empty rows much easier to handle.

Reviewed the mutable worksheet behavior, bounds, renamed exports, and round-trip coverage. I corrected the openpyxl migration example in 4126bf5: worksheet-only indexing is different from tab indexing when chartsheets are present. The pre-1.0 rename is called out in the minor changeset.

PR #163, opened 2026-09-17
fix: a t="e" or t="d" cell a current Excel writes failed the whole load

kibertoad:
## Summary

Two cell shapes a current Excel writes made `loadWorkbook` throw, and one cell
was enough to lose the whole workbook. `<c t="e"><v>#SPILL!</v></c>` threw
`OpenXmlSchemaError: worksheet: unknown error code "#SPILL!" in <c t="e">`, and
`<c t="d"><v>2024-03-14T00:00:00</v></c>` threw
`worksheet: unknown cell type t="d"`. Both now load, and `loadWorkbookStream`
answers the same bytes the same way instead of silently reading them as empty.

## Motivation

No issue: found while reading the cell reader against ST_CellType, reproduced at
32e67d0 by unzipping a fixture with `fflate`, patching
`xl/worksheets/sheet1.xml` and re-zipping.

`ERROR_CODES` in `src/utils/inference.ts` holds the eight tokens that existed
before 2018. Excel 365 also writes `#SPILL!`, `#CALC!`, `#FIELD!`, `#BLOCKED!`,
`#CONNECT!`, `#BUSY!`, `#UNKNOWN!`, `#PYTHON!` and `#EXTERNAL!`. A service that
ingests user-uploaded spreadsheets meets these constantly: one stale
dynamic-array formula anywhere in an upload failed a file with thousands of good
rows.

`d` is a legal `ST_CellType` member (ECMA-376 §18.18.11) in ISO 29500 strict,
carrying an ISO 8601 value in `<v>` instead of a serial number. The transitional
XSD vendored at `tests/conformance/schemas/transitional/sml.xsd` does not list
it, which is why `tests/conformance/` never caught the gap.

`src/streaming/read-only.ts` did the opposite of the DOM reader on both:
`case 'e'` returned `null` for an unlisted token, and the `default` arm ran the
text through `Number.parseFloat`. So the same bytes read as empty cells through
one entry point and threw through the other, and an unmodelled `t` produced a
`NaN` that failed later, at save.

## Changes

- Preserve modern and lexically safe future Excel error tokens in eager/streaming reads, writes and formula caches. Treat blank error cells as empty, and reject malformed error text on read and save.
- Share known error tokens with the formula tokenizer.
- Read ISO dates, time-only values and day/time durations from Transitional producer output. Validate calendar components, timezone offsets, end-of-day fractions and safe duration ranges.
- Normalize ISO formula caches to numeric Excel serials using the workbook's 1900/1904 epoch, so saving does not silently change a date cache into text.
- Reject unknown formula cell types and malformed date caches during loading.
- Keep the Strict package normalization contract introduced in 0.20.0: supported calendar dates become numeric serials; time-only/duration and sub-millisecond dates retain the existing explicit refusal.
- Document these semantics and include a minor changeset, remaining below v1.

## Testing

- Maintainer-added regressions reproduced 18 failures before fixes, covering invalid offsets, fractional 24:00, incomplete/overflowing durations, formula-cache save/reload under both epochs, and unknown formula types.
- All 56 focused eager/streaming cases pass. Before integrating #176, all 2,994 tests, typecheck, lint, knip, build, size, attw and consumer checks passed; all 11 performance tests passed with `PERF_GATE=1`.
- Final integration with #176: all 3,023 tests pass. Combined minified sizes are 226.18 KB (streaming) and 430.10 KB (io), so parse-size budgets increase from 225/430 to 230/435 KB to account for the two new parsers. Brotli sizes remain 65.31/111.94 KB, within unchanged 80/120 KB budgets. GitHub CI covers Linux, macOS and Windows on Node 22, 24 and 26.

## Breaking changes

`ExcelErrorCode` widens from a closed literal union to `` `#${string}` ``. Consumers with exhaustive switches must add a fallback for future error tokens. Malformed error/date values and unknown cell types now fail during loading instead of being silently dropped or failing later during save. This is released as a pre-v1 minor change.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you very much for covering modern Excel error values and keeping eager and streaming readers consistent!

I reproduced and fixed the remaining boundary problems: invalid timezone offsets, fractional end-of-day times, incomplete or overflowing durations, unknown formula cell types, and ISO formula caches becoming strings after saving. Formula date caches now retain numeric values under both workbook epochs. The Strict OOXML contract from 0.20.0 remains unchanged and is documented separately.

The combined branch passes all 3,023 tests, typecheck, lint, knip, build, size, attw and package-consumer checks. The ISO changes also pass all 11 local performance tests with PERF_GATE=1. The small parse-size budget adjustment reflects the measured parser additions; compressed-size budgets are unchanged.

PR #164, opened 2026-09-17
fix: name what the bytes were when a load fails, and document the error contract

kibertoad:
## Summary

Failed loads now add bounded hints for text/CSV, Unicode byte-order marks, PDF and ZIP prefixes. Short files report their byte length against the minimum ZIP trailer size. The README and API documentation explain the error hierarchy and recovery considerations.

## Motivation

An upload can have an .xlsx extension while containing another format or an incomplete archive. The generic invalid-ZIP message did not help callers diagnose these cases.

## Changes

- Inspect at most 64 leading bytes on failure, without changing successful ZIP reads or adding dependencies.
- Distinguish a missing EOCD signature from a detected signature; neither hint claims that the entire archive or central directory is present.
- Document error subclasses, invalid options, transient source I/O and limit failures. Exceeding a limit does not prove workbook validity. Error message wording is not a stable API.
- Add a patch changeset.

## Testing

- 16 diagnostic/class-contract tests, including malformed metadata that points at a local header instead of a central directory. The new diagnostic regression fails before the review fix.
- All 3,039 tests pass, including XML schema conformance.
- Typecheck, lint, knip, build, size, attw and packed-consumer checks pass locally.

## Breaking changes

None. Existing error classes and public signatures are unchanged; diagnostic wording changes.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you very much for making failed uploads easier to diagnose! I checked the error hierarchy and ZIP fallback paths, corrected the recovery guidance and the EOCD completeness assumption, and added a regression for a missing central directory. All 3,039 local tests and the packaging checks pass.

PR #165, opened 2026-09-17
feat: name ISO 29500 strict instead of failing with a missing-relationship error

kibertoad:
## Summary

Workbooks saved as “Strict Open XML Spreadsheet” now fail with `OpenXmlNotImplementedError` naming the format and explaining how to re-save as “Excel Workbook (.xlsx)”. Previously, they produced a missing-relationship error or, when only the workbook part used Strict namespaces, silently loaded with no sheets. Actual Strict reading remains tracked in #166.

## Motivation

Strict and Transitional workbooks both use `.xlsx` and the OPC package layer, so upload services need an actionable diagnosis from the loader. No separate issue was opened for this diagnosis; #166 tracks the remaining reading capability.

## Changes

- Detect Strict relationship types and Strict roots in the existing eager and streaming loaders and part readers through an internal helper.
- Require the workbook root to use the SpreadsheetML namespace; unrelated namespaces no longer load as empty workbooks.
- Check the worksheet root when streaming rows, including bounded row reads. A Strict worksheet behind Transitional workbook metadata is rejected consistently instead of receiving partial, unsupported treatment.
- Document the error and re-save guidance. No new public exports; writing remains Transitional.

The detector matches the Strict markup family, not arbitrary `purl.org` URIs. OPC namespaces and Dublin Core metadata remain unaffected. Tests made by rewriting namespace URIs exercise detection only; they do not establish Strict interoperability.

## Testing

- `pnpm vitest run tests/io/iso-strict-package.test.ts`: 8 cases covering fully Strict packages, mixed parts, both loaders, full and bounded streaming rows, unrelated workbook namespaces, and unchanged OPC/Dublin Core handling.
- Reproduced the missing streaming worksheet-root check before adding the regression assertions and fix.
- The combined release candidate passes all 2,867 tests, typecheck, lint, knip, build, bundle limits, and the performance suite.

## Breaking changes

Strict input now throws `OpenXmlNotImplementedError` instead of a misleading schema error or an incomplete read. Workbook roots in unrelated namespaces now throw `OpenXmlSchemaError` instead of returning an empty workbook. No export changes; documented in a minor changeset while the package remains pre-1.0.

## Checklist


baseballyama (approved):
Thanks for making this failure actionable!

I reproduced one remaining gap: a Strict worksheet behind Transitional workbook metadata was still accepted by streaming row iteration. Fixed in 563e166, with regression coverage for full and bounded reads. Both loaders now report the unsupported format consistently for that case; actual Strict interoperability remains tracked separately in #166.

PR #167, opened 2026-09-17
feat(styles): read a cell as the text Excel shows, and as a Date

kibertoad:
## Summary

Adds `getCellDisplayText(wb, cell)` and `getCellDate(wb, cell)` to
`@office-kit/xlsx/styles`. The first renders a cell's value through the number
format its style points at, so reading a workbook back gives the text a person
sees in Excel. The second reads a date-formatted serial as a `Date`, which until
now took three calls the docs never put together.

## Motivation

Opened without an issue: the gap turned up while reading workbooks produced
elsewhere, and it is a missing capability rather than a bug report to reproduce.

There was no way to ask what a cell says. The only string-producing helper,
`cellValueAsString`, takes a `CellValue`, so it never sees the stylesheet:

| Cell                              | Excel shows    | `cellValueAsString`  |
| --------------------------------- | -------------- | -------------------- |
| `0.5` with numFmt `0.0%`          | `50.0%`        | `0.5`                |
| `1234567.891` with `#,##0.00`     | `1,234,567.89` | `1234567.891`        |
| `=1.1+2.2` with `General`         | `3.3`          | `3.3000000000000003` |
| serial `45365` with `yyyy-mm-dd`  | `2024-03-14`   | `45365`              |
| `true`                            | `TRUE`         | `true`               |

The date row is the harmful one. Excel stores a date as a day count, so the
number format is the only evidence that `45365` means 2024-03-14. Before this
PR, `isDateFormat` had no caller on the read path, and a consumer that treats
cells as text stored the number. The recovery path existed but was neither
composed nor documented: `getCellNumberFormat`, then `isDateFormat`, then
`excelToDate` with the epoch taken from `wb.date1904`. `cellValueAsDate`
deliberately does not do it and says so in its docstring.

**This is not a parallel path to `cellValueAsString`.** The two take different
inputs and answer different questions. `cellValueAsString(value)` answers "what
is this value in JavaScript terms" and is correct for JSON, arithmetic, and
anything typed. `getCellDisplayText(wb, cell)` needs the workbook because the
format lives in the stylesheet, and answers "what does Excel print here": CSV
export, an HTML table, a workbook diff by what it shows, a log line for a human.
Neither can be derived from the other, and choosing wrong fails silently, so the
docstring on each names the other and says which one the reader wants.

`getCellDate` is the same argument one level down: `cellValueAsDate` takes a
value and cannot apply the serial conversion, which is exactly why its docstring
already points at `excelToDate`. Graduating the three-step composition into one
documented call is the `CLAUDE.md` "reachable but only at an abstraction every
real user re-implements" case.

## Changes

- `getCellDisplayText(wb, cell): string` in `@office-kit/xlsx/styles`. Covers
  `General` at Excel's 15 significant digits (this is what removes the float
  noise), every code in `BUILTIN_FORMATS` (one caveat below), and the common
  custom codes: `0` `#`
  `?` digit placeholders, thousands grouping, trailing commas that divide by a
  thousand, `%`, currency and quoted or `\`-escaped literals, `_x` width
  reservations, colour codes, `E+`/`E-` including the engineering step of
  `##0.0E+0`, `?/?` fractions with a placeholder or a spelled-out denominator,
  date and time codes with month and weekday names and `AM/PM`, the elapsed
  forms `[h]` `[mm]` `[ss]`, and the positive / negative / zero / text sections
  of a multi-section code. Booleans render `TRUE` / `FALSE`, errors their token,
  rich text its runs joined, a formula cell its cached value, an empty cell `''`.
- `getCellDate(wb, cell): Date | undefined`. Reads the cell's format with the
  same parser `getCellDisplayText` renders with, requires a part that names a
  day (`y`, `m` or `d`), and converts under `wb.date1904`. A `Date` value passes
  through. `undefined` covers a number whose format says it is a count, a time
  of day (`h:mm`, which names a moment inside a day but not which day), an
  elapsed span (`[h]:mm`), and a code outside the renderer's subset. One parser
  for both readings is what keeps them from disagreeing about a cell: a
  regex-based date check says `[DBNum1]yyyy-mm-dd` is a date while the renderer
  prints its bare serial.
- `docs/migrate-from-sheetjs.md`, with `codepage` and `cellDates` answered as
  "not applicable" and the reason, and `.csv` / `.xls` / `.xlsb` / `.ods` as out
  of scope with a pointer.
- A "Read a workbook somebody else produced" recipe, and a short SheetJS
  subsection in the README pointing at the guide.
- Minor changeset.

### Where the boundary is, and why it is a fallback rather than a guess

A code outside the implemented subset returns from the parser as `undefined` and
the cell falls back to `cellValueAsString`, so the reading degrades to a plain
coercion instead of printing something invented. That covers comparison sections
(`[>=100]"over";0`), era and calendar tokens (`g`, `e`, `b`, `[DBNum1]`), a
section splicing two numeric layouts together (`0.00" ("0.00")"`), more than
four sections, and an unterminated `"` or `[`. A serial past the range a `Date`
covers degrades the same way rather than printing NaN fields. The docstring
lists all of them.

Three more cases are judgement calls, because Excel's answer is not recoverable
from the cell, and the docstring says so rather than implying precision that is
not there:

- `*x` repeats `x` until the column is full. Nothing is printed for it.
- Month and weekday names come out in English; a `[$-409]`-style tag is read for
  its currency symbol and otherwise ignored.
- Column width decides how many digits `General` shows and when Excel switches
  to scientific notation. Every stored digit is kept.

### Decisions a reviewer may want to push back on

- **Named `getCellDisplayText`, not `formatCellText`.** In this codebase a
  `format*` name is a style mutator (`formatAsHeader`), so `formatCellText`
  would read as one. `getCell*` is the established prefix for "resolve this
  cell against the stylesheet and hand me the answer", next to
  `getCellNumberFormat` and `getCellFont`, and "display text" is Excel's own
  vocabulary.
- **A `Date` value under a non-date format prints the serial** (`45365`, not an
  ISO string). That is what the file does and what Excel shows: the worksheet
  stores a serial and the style decides what it means. Printing a date under a
  `General` cell would make the helper disagree with the file it describes.
  `setCellAsDate` is the write-side answer and the docstring points at it.
- **A negative serial under a calendar date code has no reading.** Excel fills
  the cell with `#` characters and how many depends on the column width, so the
  value falls back to its General text. An elapsed code is the exception: a span
  has a sign, `-1:30` is representable, and a duration value is the obvious way
  to reach one.
- **A negative value that rounds to zero prints without the sign** (`0.00`, not
  `-0.00`). Excel's own choice here could not be verified; a minus in front of an
  all-zero rendering is the worse of the two to guess.
- **The format parse is memoised** in a module-level `Map` capped at 256
  entries. Distinct codes in a workbook number in the tens while cells number in
  the millions, and the whole point of this call is bulk reads. The keys arrive
  from an untrusted `styles.xml`, so a full map evicts its oldest entry rather
  than clearing, and a code longer than the 255 characters Excel allows is
  parsed on each call instead of taking a slot.

### Risk

All of it is additive and read-only: no existing export changed, and nothing in
the `io` or `streaming` graphs imports the new modules. The one module they do
share is `numbers.ts`, which the parser reads the bracket colour names from
rather than listing them a second time; the names are stored in the regex
alternation form `STRIP_RE` already needed, which keeps `/io` byte-identical
(the array form that reads better cost it 0.14 kB brotlied). The risk that is
worth review attention is rendering fidelity, not breakage: a format code that
renders subtly wrong is worse than one that degrades, so the boundary above is
the part to argue with.

### Left out deliberately

- **No `LoadOptions.cellDates`.** It would make the value model depend on a load
  option, so every downstream `typeof cell.value === 'number'` becomes
  conditional on how the file was opened.
- **No streaming equivalent.** `ReadOnlyCell` carries a `styleId` but
  `loadWorkbookStream` exposes no `Workbook`, so there is no stylesheet to
  resolve against. Worth a follow-up if somebody asks.
- **Fraction padding beyond the mainline.** `# ??/??` pads the numerator and
  denominator so the slash lines up, which is Excel's documented behaviour. Two
  details are reasoned from the placeholder semantics rather than verified
  against Excel: blanking the fraction of an exact whole number, including a
  spelled-out denominator, and filling a `0`-placeholder denominator from the
  right (`# ??/00` on 2.5 gives `2  1/02`), since a pad zero on the right would
  multiply the denominator by ten.

### One thing the sweep turned up, not fixed here

`tests/styles/display-text-builtins.test.ts` walks all 36 codes in
`BUILTIN_FORMATS`. Every one renders as Excel does except the entry for
numFmtId 44, which mirrors openpyxl in dropping the `;` section separators
Excel's own accounting code carries
(`_("$"* #,##0.00_);_("$"* \(#,##0.00\);_("$"* "-"??_);_(@_)`). The stored code
is therefore a single section holding both digit placeholders and `@`, so it
reads as a text layout and falls back to General. That affects anything reading
`getCellNumberFormat` on a cell whose xf points at numFmtId 44 with no explicit
`numFmt` override, not only this PR. Changing the constant is a change to an
existing public export and wants its own changeset, so the sweep pins the
current behaviour with a comment and the docstring says so rather than
overclaiming. Happy to open the follow-up.

## Testing

- `pnpm test` (`tests/styles/display-text.test.ts` is the table; the other three
  new files are the date helper, the properties, and the read path).
- `tests/styles/display-text-read-path.test.ts` writes real bytes, loads them
  back through `loadWorkbook`, and asserts the rendered text, so the number
  formats have to survive the stylesheet round-trip. One case asserts both
  readings side by side (`50.0%` against `0.5`) so the two helpers cannot drift
  into agreement.
- The property tests pin four invariants: a fixed-decimal code prints exactly
  its decimals and parses back within half of the last place, `#,##0` groups
  without gaining or losing a digit, a rendered fraction reads back to within
  half a step of the denominator the code allows whatever its placeholders are,
  and the text a date code prints matches the fields of the `Date` `getCellDate`
  hands back for the same cell under either epoch.
- `pnpm typecheck`, `pnpm lint`, `pnpm knip`, `pnpm build`, `pnpm size` all pass.
  `pnpm size` shows `/styles` at 14.38 kB against a 60 kB budget (9.2 kB
  before), `/io` at 99.28 kB and `/streaming` at 60.62 kB.

Reproduce the reading this PR adds:

```sh
pnpm vitest run tests/styles/display-text.test.ts tests/styles/display-text-read-path.test.ts
```

Locally, 39 tests fail in `tests/conformance/` and in the `validateXlsx` case of
five `issue-*` files, all with `xmllint not available: spawnSync xmllint ENOENT`.
They are unrelated to this change and pass in CI, which installs
`libxml2-utils`.

## Breaking changes

None. Two new exports, no change to an existing signature or behaviour.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you very much!! The format-aware text and date readers are useful additions, and the documented fallback boundaries help keep their behavior understandable.

I added regressions and fixes in b516246 for literal-only numeric/text sections, fixed denominators containing zero (including literal suffixes), overflow during percent scaling, and date detection using the section selected for the actual value. The two focused display/date suites pass all 145 tests, and the combined candidate passes the full suite and packaging checks.

PR #168, opened 2026-09-17
perf(reader): walk <sheetData> with SAX instead of building a node per cell

kibertoad:
## Summary

`loadWorkbook` now reads worksheet cells with a synchronous SAX pass instead of allocating an XML node for every `<c>` and `<v>`. Other worksheet metadata keeps its existing tree reader. This reduces the large-sheet allocation cost while preserving namespace-aware cell and rich-text interpretation.

## Motivation

Profiling large worksheets showed that building the cell XML tree dominated load time and transient memory. This PR came from profiling the reader; there is no separate issue.

## Changes

- Parse only the worksheet header to locate the direct SpreadsheetML `<sheetData>` child and capture its inherited namespace bindings. A similarly named element inside an extension cannot replace the actual cell span.
- Walk the cell span with namespace resolution enabled. Foreign elements retain their namespaces, unbound prefixes fail, and inline rich text reuses the existing rich-string reader.
- Keep a single eager cell-reading implementation. Feed SAX in chunks to avoid making another full-sized copy of the cell span.
- Preserve DTD rejection and reject malformed XML and mixed content in the cell span with `OpenXmlSchemaError`.
- Consolidate Clark-name extraction as the internal `localNameOf` helper.
- Add a read-throughput gate alongside the existing write and memory gates. No public API changes and no bundle-budget increase.

The surrounding worksheet metadata continues to use the existing tree parser. This change does not make every part reader strict about damaged XML. The asynchronous streaming reader retains its separate iteration model; existing cross-reader tests pin row/cell coordinate agreement.

## Testing

- `pnpm vitest run tests/worksheet/sheet-data-sax.test.ts`: 46 cases, including namespace decoys, foreign value/rich-text elements, unbound prefixes, quoted markup, malformed XML, and mixed content.
- `pnpm vitest run tests/worksheet/row-without-r-attribute.test.ts` checks agreement with streaming coordinate placement.
- Review reproduced wrong values from namespace-blind matching and ignored mixed content before adding fixes and regression tests.
- Combined release candidate: all 2,867 tests pass, including conformance checks with `xmllint`; typecheck, lint, knip, build, and size checks pass.
- All 11 tests in the performance suite pass. On the review machine, the 300,000-cell read gate reaches approximately 396,000 cells/s against the 120,000 cells/s floor.
- Combined candidate `io` bundle: 107.42 kB minified+brotli (120 kB limit), 415.81 kB minified parse size (420 kB limit). The remaining parse-size headroom is small; the budget stays unchanged.

The throughput floor accounts for shared CI runner variability. Performance measurements are machine-dependent; CI checks the final pushed commits independently.

## Breaking changes

No public API change. Malformed XML inside `<sheetData>` that the previous tolerant tree parser recovered from now throws `OpenXmlSchemaError`; examples include unclosed rows, stray closing tags, undefined entities, and element content inside `<v>` or `<f>`. This is documented in the patch changeset. Valid SpreadsheetML namespace handling is preserved.

## Checklist


baseballyama (approved):
Many thanks for the profiling and performance work!

Review reproduced incorrect values when an extension supplied a competing sheetData element or a foreign-namespace value, plus unbound-prefix and mixed-content cases that were silently accepted. Fixed in b1df05d: the locator resolves the direct child and inherited namespaces, and the SAX walk preserves namespace and mixed-content validation. The expanded 46-case suite passes, along with the full candidate suite and all performance gates. The bundle limits remain unchanged.

PR #169, opened 2026-09-17
feat(io): cap the cells and rows a load will model

kibertoad:
## Summary

`LoadOptions` could bound the bytes an archive inflates to and nothing else, so
a caller had no way to say "nothing over N cells". `contentLimits: { maxCells,
maxRows }` adds that cap to `loadWorkbook` and `loadWorkbookStream`, charged
inside the cell loop so a workbook past the cap is refused before its model is
built. Unlimited by default.

## Motivation

No issue: found while sizing the defaults for an ingestion path.
`decompressionLimits` defaults to 512 MiB per entry, 1 GiB total and a
compression ratio of 1000. Those bound inflated bytes, which is not the quantity
that decides what a read costs. Cells are.

A 2 MB upload that inflates to a few hundred MB of `<sheetData>` sits inside
every one of those defaults, and until this change the model was built for every
cell of it before the caller got a chance to look at anything. A service that
accepts spreadsheets from strangers could bound the archive and still be handed
a multi-minute, multi-gigabyte load.

## Changes

- `LoadOptions.contentLimits` and `LoadWorkbookStreamOptions.contentLimits`,
  both `{ maxCells?, maxRows? }`, both unlimited when absent.
- `OpenXmlContentLimitError`, exported from `@office-kit/xlsx/utils`. Names the
  cap and the cell or row that reached it:
  `worksheet: reading Data!D10 passes contentLimits.maxCells of 39`.
- `ContentLimits`, exported from `@office-kit/xlsx/worksheet`, which is where
  it is enforced. This mirrors `DecompressionLimits` living in
  `@office-kit/xlsx/zip` while being named in `LoadOptions`.
- A cap that cannot mean anything (`0`, negative, fractional, `NaN`,
  `Infinity`) is rejected from the `loadWorkbook` / `loadWorkbookStream` call
  that carried it rather than at the first cell, with an `OpenXmlError`.
- A streaming pass is charged for the rows it walks, not the ones it yields,
  and the row-offset index a band query builds is charged for the rows it
  records. Without that, `iterRows({ minRow: 2 })` to skip a header row paid
  the whole inflate-and-index cost before the first charge, which is the case
  an ingestion service is most likely to be in.
- `SECURITY.md`: a `contentLimits` bullet in the hardening section, and the
  out-of-scope DoS bullet now names both options.
- `docs/migrate-from-openpyxl.md`: the two places that mention
  `decompressionLimits` mention this too.

### The decisions worth arguing about

**Cells and rows, not bytes of model.** A byte budget on the model would need a
size estimate per cell that is wrong the moment the value shape changes, and a
caller cannot reason about it. Cell count is what a caller can actually put in a
policy, it is what the file itself advertises, and cost is close to linear in
it.

**What it bounds, and what it does not.** The cap bounds the time a read
spends modelling cells and the size of the model it hands back. It is not a
memory limit: `loadWorkbook` inflates a worksheet part and decodes it to a
string before the first cell of it is charged, so the bytes in flight stay
bounded by `decompressionLimits` alone. The docs, the changeset and
`SECURITY.md` say that rather than promising a heap ceiling the option cannot
deliver.

**A band query is charged for the rows it has to reach.** `iterRows({ minRow })`
either walks everything before `minRow` or indexes the whole part to seek past
it, holding one `{row, offset}` object per row of the sheet while it does.
Counting only the rows a band yields left that unbounded: measured on a
500k-row sheet, a `minRow: 2` refusal took 438 ms against 10 ms for the
whole-sheet path. Both paths now charge what they read. The index is built once
per worksheet handle and charged once, against a budget of its own, so a repeat
band query is not charged twice for it.

**Charged before the cell is built.** The whole value of the option is that a
refusal is cheap. `chargeCell` runs ahead of `readCell`, so the load stops after
reading the cells up to the cap instead of modelling the sheet and then
reporting that it was too big. A test pins the shape of that cost: a 200 000-cell
workbook with `maxCells: 10` is refused well inside 2 seconds, against the
seconds and hundreds of MB modelling it takes.

**One budget for the whole `loadWorkbook` call.** A per-sheet cap would let a
workbook of a thousand small sheets straight through a ceiling that exists to
bound the read. So `maxCells: 40` against two 30-cell sheets fails partway
through the second one, and the message says so.

**`loadWorkbookStream` counts per traversal, not per workbook.** That reader
holds one row at a time, so the cap bounds how long a pass can run rather than
how much it retains. A cumulative total would refuse the second iteration of a
sheet whose first iteration it had allowed, which is a worse surprise than the
inconsistency. Both readers are documented as "one pass over the content", and
the row-iteration is that reader's pass.

**Default unlimited.** A default that refused any workbook would be a silent
breaking change for every existing caller, and there is no cell count that is
wrong for everyone. `decompressionLimits` can have a default because a 1 GiB
archive really is pathological for everybody; 500 000 cells is a Tuesday for
some callers and a DoS for others. `SECURITY.md` says this in as many words,
next to the profile to start from.

**`OpenXmlContentLimitError` extends `OpenXmlError` directly**, not
`OpenXmlIoError` the way `OpenXmlDecompressionBombError` does. Nothing failed at
the I/O layer: the file is fine and larger than allowed. The bomb error is under
`OpenXmlIoError` to keep older `catch` sites working, which is history a new
error does not have to inherit. A caller wants to tell "too big" from "corrupt"
to answer an upload with the right status.

**Left out:** no cap on sheet count, string-table size or style count. Each is a
real axis, none of them is the axis that dominates, and adding knobs nobody has
asked for is the wrong way to find out which ones matter. There is also no
`maxCells` equivalent on the write path; a caller building a workbook already
knows how big it is.

## Testing

`tests/io/content-limits.test.ts`, 18 cases: both entry points, both caps, the
workbook-wide count across two sheets, the per-traversal count and a repeated
pass, the indexed band-query route into the streaming iterator, the rejection of
a nonsense cap from both calls (including the band query that yields nothing and
so never reaches a row), the rows an index or a walk is charged for on a band
query, the same on a sheet whose rows carry no `@r` and cannot be indexed, the
unlimited default, and the cost shape of an early refusal.

```sh
pnpm vitest run tests/io/content-limits.test.ts
```

Fail-before, pass-after, verified by restoring `src/` to the base commit and
re-running: 13 of the 18 fail, 5 pass (the cases that assert an unlimited read
still works, which is the behaviour this keeps). The 5 cases added for the
review fixes were run the same way against the first commit of this branch:
all 5 fail there and pass now.

CI is green on the full matrix (3 OSes x Node 22/24/26), which is where the XSD
tier actually runs; `libxml2-utils` is not installed on my machine, so the
conformance tier is the one thing I did not verify locally. `pnpm typecheck`,
`pnpm lint`, `pnpm knip`, `pnpm build`, `pnpm size` and `PERF_GATE=1
pnpm test:perf` pass locally.

The per-cell charge is an increment and a compare against an `Infinity` cap
when no limit is set, and the index scan now pays the same per row.

Size against `main`: `io` min+brotli 107.42 kB to 107.68 kB of 120 kB, `io`
parse size 415.81 kB to 416.82 kB of 420 kB, `streaming` min+brotli 60.97 kB to
61.19 kB of 80 kB, `streaming` parse size 212.27 kB to 213.26 kB of 220 kB. The
`io` parse-size headroom is 3.2 kB and this spends 1 kB of it.

## Breaking changes

None. The option is absent by default and an absent option reads exactly as
before.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thank you for adding these ingestion limits!

I fixed one gap found during review: rows without `r` could accumulate raw/pending cells before either budget was charged. Both readers now charge rows at the opening tag and cells before buffering their content, including unresolved rows that might later fall outside a streaming band. The regression tests cover early refusal, exact-limit coordinate resolution, and filtered rows/columns; the streaming documentation explains what is counted.

Validation: all 2,893 tests, all 11 performance tests with `PERF_GATE=1`, lint, TypeScript, build, Knip, bundle-size limits, ATTW and the packed-package consumer checks passed locally. The changeset remains minor.

PR #176, opened 2026-09-20
fix: a legacy .xls was reported as an encrypted xlsx

kibertoad:
## Summary

Passing a legacy `.xls` to `loadWorkbook` failed with `Encrypted xlsx is not supported. Decrypt with msoffcrypto-tool first.` The file has no password. `openZip` now reads the OLE container's directory and picks the error message from what the container holds, so a `.xls` is reported as a legacy workbook that has to be converted.

## Motivation

`openZip` treated the CFB magic bytes as proof of an encrypted xlsx. A BIFF `.xls` starts with the same eight bytes, so every `.xls` got decryption advice.

Found while planning the move of [nlpjs-neo](https://github.com/kibertoad/nlpjs-neo) from SheetJS to this library. Its Excel models are `.xls` files, and after that migration the most likely user mistake is pointing `loadExcel` at an old `model.xls`. No issue was opened for it: the bug reproduces with any `.xls` and the fix is contained in the zip layer.

## Changes

- Distinguish root-level `Workbook`/`Book` streams (legacy `.xls`), `EncryptedPackage` (encrypted `.xlsx`), and unknown compound files. Embedded spreadsheets do not determine the outer format.
- Follow FAT and DIFAT chains with bounded traversal, including large files, cycles, truncated inputs, and non-zero-offset byte views. Recognize raw BIFF2–8 headers as legacy `.xls`.
- Export `OpenXmlUnsupportedFormatError` with a stable `format` discriminator. It extends `OpenXmlNotImplementedError`, preserving existing `instanceof` checks.
- Update README, API documentation and `llms.txt`, including compatibility with the newly merged Strict OOXML support.
- Include a minor changeset.

## Testing

- 29 classification regressions, including committed LibreOffice-generated `.xls`, password-protected `.xls`, and encrypted `.xlsx` fixtures.
- Maintainer verification: all 2,967 tests pass locally, including XML schema validation; typecheck, lint, knip, build, size, attw, and consumer checks pass.
- GitHub CI also validates Linux, macOS and Windows across Node 22, 24 and 26, plus packaging and performance.

## Breaking changes

Error messages now identify the actual unsupported format. Consumers should use the new `format` discriminator instead of matching message text. Existing `instanceof OpenXmlNotImplementedError` checks continue to work; the concrete error `name` is now `OpenXmlUnsupportedFormatError`.

## Checklist

- [x] If this is a breaking change, I have added a changeset / CHANGELOG entry and
      flagged it above.
- [ ] If I used an LLM to draft this PR, I have verified each change myself, the PR
      represents real work that warrants a maintainer's review, and I am willing to
      defend each line in review.


baseballyama (approved):
Thank you! Distinguishing root-level BIFF workbooks from encrypted OOXML makes the conversion advice much more useful, and the embedded-document and DIFAT regressions cover the tricky cases well.

I resolved the main-branch documentation conflicts and updated the PR description to match the final implementation. All 2,967 tests and the local type, lint, unused-code, build, size and package-consumer checks pass. The full GitHub CI matrix and performance gate also pass.

PR #182, opened 2026-09-20
fix: name the bad argument when an address helper is handed a Worksheet

kibertoad:
## Summary

Passing a Worksheet in the workbook or address slot of getCellAtAddress / setCellAtAddress now raises OpenXmlSchemaError identifying the bad argument and pointing to the existing worksheet coordinate helpers. Non-string parseSheetRange arguments are rejected without invoking string coercion.

## Motivation

JavaScript callers previously received a misleading missing-delimiter diagnostic containing [object Object], or a raw TypeError when a Worksheet occupied the workbook slot.

## Changes

- Validate the workbook shape and address type at the public boundary, sharing address resolution between the get/set helpers.
- Describe the received type without interpolating arbitrary input values.
- Keep the existing typed signatures and APIs; add a patch changeset.

## Testing

- Regression coverage for misplaced Worksheets, omitted arguments, null, primitives, arrays, symbols and a throwing toString method. JavaScript misuse is exercised through Reflect.apply without unsafe type casts.
- All 3,037 tests pass, including XML schema conformance.
- Typecheck, lint, knip, build, size, attw and packed-consumer checks pass locally.

## Breaking changes

Valid typed calls are unchanged. Invalid JavaScript calls now consistently throw OpenXmlSchemaError; some previously escaped as native TypeError or were coerced into strings. Diagnostic wording also changes.

## Checklist

      the result.
      marked accordingly.
      comments.


baseballyama (approved):
Thanks for improving these diagnostics! Both misplaced-Worksheet cases now identify the bad argument, and the added JavaScript-boundary tests cover symbols and throwing string coercion without unsafe casts. Valid address calls remain unchanged; all 3,037 tests and the local packaging checks pass.

PR #184, opened 2026-09-22
fix: cell text shaped like _xHHHH_ lost characters on save

kibertoad:
## Summary

A cell whose text already looks like the `_xHHHH_` escape convention loses characters on save. `escapeCellString` protects such text by escaping the underscore that opens a sequence, and it misses two shapes of opener, so a reader decodes text the user wrote literally. `_x0041_x0042_` comes back as `_x0041B`, and `"SKU_x0041\nrest"` comes back as `SKUAx000A_rest`.

## Motivation

Opened without an issue: found while auditing the escape layer, not reported by a user.

The convention is that `_` is a literal in normal text but a sequence opener in escape position, so an existing `_xHHHH_` is stored as `_x005F_xHHHH_`. Whether an underscore is an opener depends on what closes the sequence, and there are two sources for that character.

The closer can be shared. In `_x0041_x0042_` the underscore at index 6 closes the first sequence and opens the second, and `String.prototype.replace` does not overlap matches, so a pattern that consumes the closer protects one opener and leaves the next one exposed.

The closer can also be absent from the input entirely. An illegal codepoint is written as `_xHHHH_`, which puts an underscore straight after whatever precedes it: `SKU_x0041` followed by a line break goes out as `SKU_x0041_x000A_`. Excel decodes left to right without overlapping, the same rule our reader applies, so the file on disk is genuinely wrong rather than merely misread here.

| written | read back |
| --- | --- |
| `_x0041_x0042_` | `_x0041B` |
| `a_x0044_x0045_x0046_b` | `a_x0044Ex005FFb` |
| `SKU_x0041` + `\n` + `rest` | `SKUAx000A_rest` |
| `col_x0009` + `\t` + `val` | `col` + `\t` + `x0009_val` |

## Changes

- `escapeCellString` matches the opening underscore alone, through a lookahead built from the same character class the illegal-codepoint pass replaces. Either pass can supply the closer, and the closer stays unconsumed so it can open the sequence that follows. Text of this shape round-trips through shared strings, rich-text runs and inline strings.
- `loadWorkbook` decodes each `<t>` of a shared string on its own rather than concatenating the raw text and decoding once. Two halves could combine into a sequence neither of them held, and `loadWorkbook` and `loadWorkbookStream` returned different text for the same bytes; the streaming reader already decoded per element. The `<r>` branch of the same helper is unreachable (`parseRichString` returns before it whenever a run exists) and goes with it.
- No change for input with a single sequence: `foo_x0041_bar` still emits `foo_x005F_x0041_bar`, so existing files and existing golden bytes are unaffected.

## Testing

```sh
pnpm vitest run tests/utils tests/io tests/workbook/shared-strings.test.ts
```

- `tests/utils/escape.property.test.ts` pins the invariant rather than samples of it: escape followed by unescape is the identity. The generator's alphabet is the convention's own characters, because a general string generator produces `_x0041_` about never. It fails against `main`.
- The inline-string case runs through the streaming writer, the only path that emits `<is>`: the test fills the shared-string table past its payload budget first, then reads the samples back through `loadWorkbookStream` and through `loadWorkbook`.
- Full suite: 42 failures here and 42 on `main`, all of them the conformance tiers that shell out to `xmllint`, which is absent on this machine. CI installs it.

## Breaking changes

None. The stored form changes only for text that previously round-tripped incorrectly; anything that round-tripped before is byte-identical.

## Checklist


baseballyama (approved):
Thank you so much! Matching the opener alone so the closer stays free to open the next sequence is the right fix, and the property test over the convention's own alphabet is what makes it convincing. Good call dropping the dead `<r>` branch in `collectText` while you were there.

PR #185, opened 2026-09-22
fix: copyRange corrupted an overlapping same-sheet copy

kibertoad:
## Summary

`copyRange` corrupted an overlapping same-sheet copy, and both range movers left stale cells sitting inside the block they had just written. `copyRange` now reads the source before writing any of it, and both functions replace the landing rectangle whole rather than stamping populated cells over whatever was already there.

## Motivation

Opened without an issue: found while auditing the range helpers, not reported by a user.

Shifting a block by a row or a column is the ordinary way to make space in a table, and it is exactly the case that broke. `copyRange` read `ws.rows` inside its write loop, so `A2` was written (as the copy of `A1`) before it was read (as the source for `A3`), and `[1, 2, 3]` copied from `A1:A3` to `A2:A4` collapsed to `[1, 1, 1, 1]`.

Reading the whole source up front fixes a dense range but not a sparse one. With `A2` empty, the same call writes `A2` and `A4` and never touches `A3`, so the old `3` survives inside the pasted block: `[1, 1, 3, 3]` where Excel gives `[1, 1, _, 3]`. The gaps have to travel with the values.

That raises the wider question of whether an empty source cell should blank its destination at all. Answering it only for the overlap leaves the library deciding the same question two ways depending on whether two rectangles happen to touch. Replacing the landing rectangle whole is one rule, it is what pasting a selection means in Excel, and it removes a second defect: `setCell` writes through an existing cell, so a destination kept its own `hyperlinkId` / `commentId` when a copy landed on it, and copying an unlinked cell over a linked one left the new value wearing the old link. Clearing the rectangle first drops those cells outright, so there is nothing left to inherit.

`styleId` has a related problem that this PR only documents. It indexes the source workbook's `cellXfs`, and a `Worksheet` holds no reference to its workbook, so a copy into a sheet belonging to a different workbook cannot be detected here, let alone remapped. `targetWs` being same-workbook is now a stated precondition on both functions instead of an unwritten one.

## Changes

- `copyRange` / `moveRange` replace the landing rectangle whole. A coordinate whose source cell is empty now lands empty instead of keeping the destination's own value.
- A destination cell no longer keeps its `hyperlinkId` / `commentId` when a copy or move lands on it. Cross-sheet copies still drop the source's ids, unchanged.
- The returned count covers populated source cells only. For a sparse source overlapping its target it can now be lower than before: the old count included destination cells the write loop had created itself and then re-read.
- Both docblocks state the same-workbook requirement on `opts.targetWs`.

## Testing

```sh
pnpm vitest run tests/worksheet/copy-range.test.ts tests/worksheet/move-range.test.ts
```

Ten new cases. The three that pin something a reviewer would otherwise have to take on trust: a sparse source over an overlapping target, which is the shape the read-first fix alone does not handle; a copy landing on a cell that already carries a comment; and a cross-sheet copy asserting both sheets, since the read side and the write side are different worksheets there and swapping them shows up nowhere else.

Full suite: 42 failures on this branch and 42 on `main`, all of them conformance tests that need `xmllint`, which is absent on this machine.

## Breaking changes

Behaviour, not signatures. The changeset is `minor`.

Callers using `copyRange` / `moveRange` as a merge, where populated cells are stamped over the destination and gaps are left alone, get a paste instead. A copy from a range with holes blanks the matching destination coordinates, and a destination cell loses a hyperlink or comment of its own. The merge is still reachable directly: iterate `getCellsInRange` over the source and `setCell` each one at its offset.

## Checklist


baseballyama (approved):
Thank you so much! Reading the source band before the first write is the right fix, and deriving the landing rectangle from one `copiedExtent` calculation keeps the cells read and the cells replaced from drifting apart. Replacing the whole rectangle is a behaviour change, but it matches Excel's paste and the changeset spells it out.

PR #186, opened 2026-09-22
fix: an edit to one column dropped its neighbours' widths

kibertoad:
## Summary

Editing one column's width on a loaded workbook silently dropped every other column's width. The reader keeps one entry per `<col>` element, so Excel's sheet-wide `<col min="1" max="16384" width="12"/>` arrives as a single run, and `setColumnDimension` deleted any run that straddled the column it was asked about. Runs are now split around the columns being edited, and every bulk helper rewrites its whole band in one pass over the runs instead of one map scan per column.

## Motivation

Opened without an issue: found while auditing the dimension helpers, not reported by a user.

Load, adjust one column, save is the most ordinary thing a caller does with an existing workbook, and it lost data. `setColumnDimension` carried a comment saying runs were dropped wholesale as "phase-5 minimum scope", but two other comments in `dimensions.ts` claimed the opposite, which is presumably why nobody noticed the combination was lossy:

- "read keeps each column its own entry so user mutations don't surprise neighbours". The reader does the reverse: `reader.ts` preserves runs verbatim, one entry per `<col>`.
- "`columnDimensions` Map keys by `min` so per-column edits stay O(log n)". `getColumnDimension` and `setColumnDimension` both scan every run, so a per-column edit is O(runs).

The second claim hid a quadratic: the bulk helpers looped over the band calling both per column, against a map that grew as they went. Measured on this machine, `hideColumns(ws, 1, 4000)` took 67 ms and `hideColumns(ws, 1, 8000)` took 271 ms, a ratio of 4.07 where linear work would give 2.

The "drop the entry when its last field goes" branches in `unhideColumn`, `ungroupColumns` and `expandColumnGroup` had the same data loss in a sharper form: they ran `ws.columnDimensions.delete(existing.min)`, which removes the whole run by its key even when the caller named one column in the middle of it.

## Changes

- An edit to one column inside a multi-column run splits the run: the column gets its own entry carrying the run's fields plus the change, and the columns on either side keep theirs under their own keys. Applies to `setColumnDimension`, `setColumnWidth`, `hideColumn`, `unhideColumn` and every bulk helper.
- `hideColumns`, `unhideColumns`, `setColumnWidths`, `autofitColumns`, `groupColumns`, `ungroupColumns`, `collapseColumnGroup` and `expandColumnGroup` rewrite their whole band in one pass. Their observable results are unchanged apart from the split above.
- A band that runs past the last column is rejected before the column list is built, so `hideColumns(ws, 1, 50_000_000)` throws `OpenXmlSchemaError` immediately instead of spending seconds building 50 million entries first.
- `unhideColumn` rejects a column outside the sheet with an `OpenXmlSchemaError`, the way `hideColumn` already did. It used to return silently.
- `dimensions.ts` describes what the reader, the writer and the lookup actually do, including that neither side merges or splits runs on its own.
- `getColumnDimension` is still a scan over the runs, and now says so rather than claiming O(log n).

## Testing

```sh
pnpm vitest run tests/worksheet/column-dimension-runs.test.ts
PERF_GATE=1 pnpm test:perf tests/perf/column-dimension-scaling.test.ts
```

- `tests/worksheet/column-dimension-runs.test.ts` covers splitting at the start, middle and end of a run, field-dropping edits, whole-band edits, group and collapse round-trips, a load-edit-save cycle, and the two degenerate map shapes below. 16 of its 18 cases fail on `main`.
- `tests/perf/column-dimension-scaling.test.ts` measures the ratio between a 4000-column and an 8000-column `hideColumns` rather than an absolute time, so it does not encode this machine's speed. Each side repeats the band 40 times: the whole point of the gate is that it runs under `PERF_GATE=1` in CI, and a ratio between two single-millisecond samples turns on one GC pause. On `main` the ratio is 4.07; here it is 2.1 to 2.25 across runs, against a ceiling of 3.
- Per-call times on this machine, warm: `hideColumns(ws, 1, 8000)` 271 ms to 0.8 ms, `setColumnWidths` with 4000 widths 64 ms to 1.5 ms, `autofitColumns` over 2000 populated columns 17 ms to 1.0 ms.
- Full suite: no change against `main` (the 42 failures on both sides are the conformance tests that need `xmllint`, which is absent on this machine).

## Breaking changes

None to the API shape. Two behaviour changes worth a reviewer's attention:

- `ws.columnDimensions` now holds the split pieces of a run that a previous version would have deleted, so code reading that Map directly sees more entries than before after an edit. The entries it sees for any given column are the ones that column actually had.
- `unhideColumn` throws for a column outside the sheet instead of returning silently. The changeset calls it out.

## Why the two findings are in one PR

Two separate problems, one code path. The split has to happen inside the same twenty lines that the one-pass rewrite replaces, and a one-pass bulk helper that did not split would re-introduce the data loss at full speed. Landing them separately would mean a second PR that rewrites the first one's diff, which costs a reviewer more than reading them together.

## Why `columnDimensions` stays a plain Map

The Map is public, mutable, and keyed by each entry's `min`. That is the reason a lookup is a scan and the reason a split has to decide what to do when two runs claim one column. A normalized, sorted, disjoint run list behind accessors would make the lookup O(log n) and the split trivial, but it takes the Map away from the callers who write into it today, so it belongs in a deliberate API change rather than in a data-loss fix.

What the split does instead is preserve exactly what `getColumnDimension` reported before the edit. The first entry covering a column wins, as it already did. Pieces are filed under their own `min`, and a piece whose key is already taken keeps the rest of its span instead of being dropped, so an edit inside two overlapping runs no longer loses the columns only the second run covers. Both degenerate shapes (two overlapping runs, and an entry filed under a key that is not its `min`) have a test pinning what survives.

## Checklist


baseballyama (approved):
Thank you so much! Splitting the covering run instead of deleting it is the right fix, and pairing the whole band against the runs in one pass is a good bonus. Collecting the pieces before filing them (rather than inserting mid-iteration) is a subtle trap avoided — thanks for the comment explaining it.

PR #187, opened 2026-09-22
fix: setPrintTitles broke on a sheet name with an apostrophe

kibertoad:
## Summary

`setPrintTitles` built its sheet-qualified reference with a raw `'${sheetName}'`, so an apostrophe inside the sheet name closed the quoted name early. A sheet called `Bob's Sheet` produced `'Bob's Sheet'!$1:$1`: Excel reads that as a broken defined name, and this library's own `getDefinedNameTarget` throws on it. The quoting now goes through a helper that doubles the apostrophe, giving `'Bob''s Sheet'!$1:$1`.

## Motivation

Opened without an issue: found while auditing the defined-name helpers, not reported by a user.

`validateSheetTitle` rejects a title that starts or ends with an apostrophe but allows one inside, which matches Excel. So `Bob's Sheet` is a title a caller can create through `addWorksheet`, and setting print titles on it produced a reference nothing can read:

```
getDefinedNameTarget threw: parseSheetRange: missing "!" delimiter in "'Bob's Sheet'!$1:$1"
```

Every other reference-building path in the library already goes through `formatSheetQualifiedRef`, which doubles the apostrophe. `setPrintTitles` hand-rolled the prefix instead, which is how it missed the rule. It cannot just call `formatSheetQualifiedRef`, because that emits a simple title bare (`Report!$1:$1`) and Excel writes this built-in name with the title quoted whether or not it needs to be. So the doubling moves into a `quoteSheetName` helper that both paths share.

## Changes

- `setPrintTitles` doubles an apostrophe in the sheet name. Output for a name without one is byte-identical, so existing files and tests are unaffected.
- New internal `quoteSheetName` in `src/utils/coordinate.ts`, used by `setPrintTitles` and by `formatSheetQualifiedRef`, so one place knows how Excel escapes an apostrophe in a sheet reference. Not re-exported from `@office-kit/xlsx/utils`: it is the unconditional half of `formatSheetQualifiedRef`, which stays the public way to build a reference.

## Testing

```sh
pnpm vitest run tests/workbook/defined-name-helpers.test.ts
```

- One case asserts the emitted value for `Bob's Sheet`.
- A second feeds that value back through `getDefinedNameTarget` and checks both legs resolve to the sheet and range they came from, which is what pins the output as readable rather than merely different.
- Both fail on `main`.

## Breaking changes

None. The emitted value changes only for sheet names containing an apostrophe, where the previous output was unreadable.

## Checklist


baseballyama (approved):
Thank you so much! Nice catch — `validateSheetTitle` accepts apostrophes, so this was reachable. Sharing `quoteSheetName` with `formatSheetQualifiedRef` keeps the doubling in one place.

PR #189, opened 2026-09-22
fix: a doubled formula prefix reached the file as <f>=A1</f>

kibertoad:
## Summary

Formula text that still begins with `=` after the leading one is removed is now rejected where a caller supplies it, and repaired where it comes out of a file. OOXML stores formula text without the UI's `=` (ECMA-376 §18.3.1.40) and Excel reports a file carrying one as damaged, but `normalizeFormulaText` strips exactly one, so a doubled prefix either reached the part or silently became a different formula.

## Motivation

Opened without an issue: found while auditing the formula-text layer, not reported by a user.

Two ways the doubled prefix got out on `main`, both verified against the branch point:

- A hand-built `{ kind: 'formula', t: 'normal', formula: '==A1' }` writes `<f>=A1</f>`. The model interfaces are public, so a literal skips the constructor and the serialiser's single strip is all that runs. That is the damaged-file shape.
- `makeFormula('==A1')` stores `'=A1'` on the value, and the serialiser then strips that one too, so the file gets `<f>A1</f>`. The caller's typo becomes a formula they never wrote, with nothing reported.

`bindValue` reaches both without the caller thinking about formulas at all, since it routes any string starting with `=` to the formula path.

Rejecting rather than stripping repeatedly is deliberate. `'==A1'` is invalid in Excel's formula bar too, so there is no intent to recover, and stripping to `'A1'` is the silent-guess outcome above.

The reader is the other side of that argument. It runs untrusted bytes through the same helper, and a workbook that fails to open because one `<f>`, one `<formula1>`, one `<cfRule>` formula or one `<definedName>` in it carries a stray `=` is a worse outcome than a repaired formula: the file is already outside the spec, and every reading of the expression is the same once the prefix is gone. So the policy splits by boundary rather than living in one shared helper.

## Changes

- `normalizeFormulaText(text, context)` throws `OpenXmlSchemaError` when the text still begins with `=`. The constructors (`makeFormula`, `makeArrayFormula`, `makeSharedFormula`, `makeDataTableFormula`, the `set*` forms, `bindValue`, `makeDefinedName`, `makeCfRule`, `makeDataValidation`) and the `<f>` / `<formula>` / `<formula1>` / `<formula2>` / `<definedName>` serialisers pass a context, so the message names the call (`makeFormula:`) or the element and where it sits (`worksheet: <f> at B7`). The offending text is quoted through `quoteCellText`, which bounds it: it can be a whole formula.
- `repairFormulaTextFromFile` strips every leading `=` and is what the reader uses for all five elements. Loading is unchanged from `main`, including for files that carry the doubled prefix.
- `bindValue` still treats a leading `=` as the whole formula test, so `'==A1'` throws there the way `'='` already did. Its doc says so and points at `setCellValue` for data that may legitimately start with `=`; `inferCellType`'s doc says that classifying a string as `'f'` is not a promise the constructors will accept it.
- Changeset is a `minor`: the constructors and the serialisers throw where they used to accept.

## Testing

```sh
pnpm vitest run tests/formula/leading-equals.test.ts
```

The cases live in the file that already covered this function through every caller. Beyond the rejections, they pin the three things a reviewer would otherwise have to take on trust: a file carrying the doubled prefix in `<f>`, `<formula>`, `<formula1>` or `<definedName>` still loads, the rejections name the call or the cell, and the quoted text stays bounded when the input is a 500-character formula.

Locally the full suite has 42 failures before and after this branch, all of them conformance tests that need `xmllint` (absent on this machine, installed in CI).

## Breaking changes

A new throw, so the changeset is a flagged `minor` rather than a patch. Calls that pass a doubled prefix now get an `OpenXmlSchemaError` where they used to get a file with a formula Excel refuses or a formula the caller did not write. Nothing changes for reading, for a single prefix, or for an `=` inside the expression (`'=A1=B1'`).

## Checklist


baseballyama (approved):
Thank you so much! Splitting the two boundaries — reject on the way in, repair on the way out of a file — is exactly right, and worth the extra function. Threading `context` through so the error names the call or the cell makes these genuinely debuggable.

PR #190, opened 2026-09-22
fix: a NaN width or height reached the worksheet part

kibertoad:
## Summary

A non-finite column width or row height went straight into the worksheet part, and Excel does not refuse it. It opens the file and turns `<col width="NaN"/>` into `width="0" hidden="1"`, so the column the caller sized silently disappears; `<row ht="NaN">` falls back to the default height. A size the caller passes is now refused at the setter, and the serializer refuses a non-finite one whatever route put it into the model.

## Motivation

Opened without an issue: found while auditing the dimension setters, not reported by a user.

`setDefaultColumnWidth` and `setDefaultRowHeight` already rejected a non-finite or negative size. Their per-column and per-row siblings did not, so the same mistake was caught in one place and serialised in the other. A `NaN` gets there without anyone typing it: a width computed from a measurement that divided by zero, or `Number(someMissingField)`.

The write-only path validated nothing at all. `ws.setColumnWidth(0, NaN)` produced `<col min="0" max="0" width="NaN" customWidth="1"/>`, an index outside the grid as well as an unusable double. That path formats `<col>` itself rather than going through `setColumnDimension`, so it never inherited `validateRowCol`.

## Changes

- `setColumnWidth` and `setRowHeight` throw `OpenXmlSchemaError` for a size that is not a non-negative finite number. `setDefaultColumnWidth` and `setDefaultRowHeight` had the same rule inline and now share it.
- `serializeColumnDimension` and `serializeRowDimensionAttrs` throw on a non-finite size, naming the `<col>` run or the row.
- The write-only `setColumnWidth` rejects the same sizes and a column index outside `[1, 16384]`, as `OpenXmlSchemaError`.
- `setColumnWidths` and `setRowHeights` skip a negative entry the way they already skipped a non-finite one.
- `autofitColumn` and `autofitColumns` validate `padding` / `min` / `max` before resizing anything.

## Testing

```sh
pnpm vitest run tests/worksheet/dimension-value-validation.test.ts
```

Each guard was deleted in turn to confirm the suite goes red for it: the funnel guard (2 tests), the serializer guard (2), the autofit check (1), the bulk-setter predicate (2), the streaming error type (3).

Excel behaviour was checked by opening probe workbooks through COM on Windows and reading what Excel writes back, rather than from the spec:

| authored | what Excel saves back |
| --- | --- |
| `width="12"` (control) | `width="12"` |
| `width="NaN"` | `width="0" hidden="1"` |
| `width="-1"` | `width="6552.69921875"`, its widest column |
| `width="1000"` | `width="1000"` |

No repair prompt in any of the four. That is what the "not enforced" decision below rests on, and it is why the fix is framed as a column that vanishes rather than a file Excel rejects.

Full suite: no change against `main`. The 42 failures on both sides are the conformance tests that shell out to `xmllint`, which is absent on this machine, so the XSD tier of the claim above is unverified locally and rides on CI.

## Breaking changes

None against `main` for a call that produced a usable file. A call that now throws previously wrote a column Excel hides or blows up to its maximum width.

One behaviour change worth naming: `setColumnWidths([10, -1, 12])` used to put a negative width in the model, and now skips that entry the way it has always skipped a non-numeric one.

## Where the guard sits, and why not on `set*Dimension`

`setColumnDimension` and `setRowDimension` look like the chokepoint, but they are the wrong place. Every mutator that flips a flag (`hideColumn`, `unhideColumn`, `hideRow`, `groupRows`, `groupColumns`, `ungroupColumns`, `collapseRowGroup`, `collapseColumnGroup`, `expandColumnGroup`) re-submits the existing entry through them as `{ ...existing, hidden: true }`. A part carrying `width="-1"` loads fine, so validating there makes all nine throw on a value the caller never supplied, and a file that opens in Excel stops being editable by this library.

So the rule is split by what each layer owns:

- The caller-facing setters refuse a size that is not a non-negative finite number, because that argument is the caller's own.
- The serializer refuses only a non-finite one. That is the class of value that cannot come from a file (the reader drops it) and cannot be written, so refusing it there cannot break a round-trip. It covers `ws.columnDimensions.set(...)`, which `setColumnDimension`'s docblock points callers at for range-spanning entries, and which no setter guard can cover.
- A negative size passes the serializer untouched, so a worksheet that arrives carrying one still saves.

Deliberately left alone: `pageMargins` and `defaultRowHeight` interpolate their values into the part the same way, a few lines from the code this touches. Both are reachable only through setters that already validate or through the reader, which drops a non-finite value, so neither has the hole `<col>` had. Widening the sweep to them belongs in its own PR.

## Checklist


baseballyama (approved):
Thank you so much! Refusing at the setter and keeping the serializer as a backstop for the directly-writable dimension maps is the right split, and letting a negative size round-trip while rejecting non-finite ones is a well-judged line. I pushed a merge commit resolving `worksheet.ts` against the column-run splitting that landed in #186.

@kibertoad — separately: I've invited you to the `xlsx-dev` team in the office-kit org, which carries write access to this repo. The quality of this batch of PRs has been excellent, so it'd be great to have you on board if you're up for it!

PR #191, opened 2026-09-22
feat: refuse cell text past Excel's 32767 code-unit ceiling

kibertoad:
## Summary

Excel accepts 32,767 UTF-16 code units in one cell. A longer string was written without complaint, and Excel then reported the workbook as needing repair. The writer now refuses it, so the failure lands where the caller can still shorten or split the text, and `MAX_CELL_TEXT_LENGTH` is exported so a caller can check first.

## Motivation

Opened without an issue: found while auditing what the writer validates, not reported by a user.

An over-long cell produced a valid-looking xlsx with no error. The producer finds out when Excel refuses to open the file cleanly, which is the worst place to find out: the file is already in someone else's hands, and the text that was too long is out of reach.

This limit is the sibling of one the library already enforces. `validateSheetTitles` checks Excel's 31-character ceiling on every sheet name at save time, so both the habit and the placement are established; the per-cell ceiling was missed. It sits next to `serializeNumber` rejecting a non-finite `<v>` and `escapeXmlTextVerbatim` rejecting a codepoint XML 1.0 cannot carry.

## Changes

- `serializeCell` throws `OpenXmlSchemaError` for cell text past 32,767 UTF-16 code units, naming the cell, the measured length and the ceiling. Covers plain strings, rich text and a cached formula string result. Both writers go through `serializeCell`, so the write-only path is covered too.
- Rich text counts across the cell's runs. The ceiling is on what the cell holds, so two 16,384-unit runs are over it even though neither run is.
- `appendRow` on a write-only worksheet claims its row number only once every cell in the row has serialised. A refused cell used to consume the number on the way out, leaving a gap Excel reads as an empty row for a caller that shortened the value and appended again.
- New export `MAX_CELL_TEXT_LENGTH` from `@office-kit/xlsx/utils`.

## Testing

```sh
pnpm vitest run tests/worksheet/cell-text-length.test.ts
```

Which unit Excel counts decides the whole implementation, so it was measured rather than argued from the spec. Four minimal packages, one shared string each, opened through desktop Excel 16.0 on Windows:

| cell text | characters | code units | Excel |
| --- | --- | --- | --- |
| 32,767 letters | 32,767 | 32,767 | opens |
| 16,000 emoji | 16,000 | 32,000 | opens |
| 16,383 emoji + 1 letter | 16,384 | 32,767 | opens |
| 16,384 emoji | 16,384 | 32,768 | needs repair |

The last two hold the same number of characters and differ by a single code unit, so the ceiling is counted in code units and `String.length` is that count. A character count would have written the last file and handed the user the repair dialog.

Two cases are worth a look in review because they pin behaviour the diff alone does not show:

- Load, then save. A file whose `sharedStrings.xml` holds an over-long string still loads, fails to save with the cell named, and saves once that cell is shortened.
- The write-only path appends a row that is refused and then appends again, asserting the emitted sheet numbers the rows without a gap.

`tests/streaming/bounded-strings.test.ts` used a 40,000-unit cell to drive a surrogate pair across the shared-strings chunk boundary. 24,000 units crosses the same boundary and is a cell Excel can open.

Full suite: 42 failures, all of them `validateXlsx` assertions reporting `xmllint not available` on this machine. CI installs `libxml2-utils`, so the conformance tier runs there.

## Breaking changes

Pre-1.0 behaviour change, flagged as a minor in the changeset rather than a patch.

A caller that was writing over-long cells now gets an error where it used to get a file. That file was one Excel would not open without repairing, so nothing that worked stops working, but a pipeline that ignored the problem will start failing loudly.

Reading is unchanged, so a file from a producer that does not check still loads and only fails on the way back out. There is deliberately no opt-out: the same asymmetry already exists for sheet titles, which the reader accepts and `validateSheetTitles` refuses at save time, and the recovery is the same one. Shorten the cell the error names, then save. `MAX_CELL_TEXT_LENGTH` is exported for callers who would rather check than catch; a slice has to keep a surrogate pair together, since a lone surrogate is rejected further down as unrepresentable in XML.

## Design notes

Why the check runs at serialize time rather than in `setCell`, where user values enter the model: `setCell` is also the reader's own path into a worksheet, so validating there would refuse to load files that load today. That is a larger break than refusing to write them, and serialize time is where both writers converge anyway.

Why three call sites rather than one dispatch over `CellValue`: a dispatcher would repeat the kind checks `serializeCell` already makes and reach into `FormulaValue.cachedValue`, which `serializeFormulaCell` owns. Two switches over one union that have to stay in step is the worse failure mode. The constant's doc comment names the rule for whoever adds the next text-carrying kind.

## Checklist


baseballyama (approved):
Thank you so much! Counting UTF-16 code units (not characters) is the right unit, and summing run lengths instead of joining keeps the write path allocation-free. Deferring `nextRow++` until the row has fully serialised is a nice touch — a rejected cell no longer burns a row number.

PR #193, opened 2026-09-22
fix: loadWorkbookStream flattened rich text to plain text

kibertoad:
## Summary

`loadWorkbookStream` joined a rich string's runs into plain text, so the same file read two ways gave two different cell values: `loadWorkbook` reported `{ kind: 'rich-text', runs: [...] }` and the streaming reader reported `"bold"`. Per-run formatting is now preserved on the streaming path too, for both shared strings and inline strings.

## Motivation

Opened without an issue: found while auditing the two readers against each other, not reported by a user.

`ReadOnlyCell.value` is typed `CellValue`, which includes the rich-text variant, so the type promised something the reader never produced. A streaming consumer narrowing with `isRichTextValue` got `false` for every cell, with no way to tell a genuinely plain string from a flattened one.

The library's own comments set the standard. `src/utils/cell-number.ts` exists so that "one corrupt value is answered the same way whether the caller loads a workbook or streams it", and `decodeCellValue`'s numeric branch says "the two entry points have to answer the same bytes the same way". Rich text was the one value shape where they did not.

For shared strings this was pure loss with no upside: `loadWorkbookStream` already parses `xl/sharedStrings.xml` into full entries and then spent an extra pass mapping them down to joined strings. Passing the entries through is less work and less memory.

Inline strings needed the `<is>` subtree. Rather than add a second definition of CT_Rst to the SAX loop, the subtree is collected into a small `XmlNode` and handed to the same `parseRichString` that `loadWorkbook` uses. That is the approach `loadWorkbook`'s own `<sheetData>` lexer takes, and `src/xml/parser.ts` already refers to "the `<is>` subtrees it builds itself". An `<is>` body is a handful of runs, so the tree costs nothing beside the row it belongs to.

## Changes

- Shared-string entries reach `decodeCellValue` whole instead of pre-flattened, so a `t="s"` cell pointing at a rich entry yields runs.
- The `<is>` body of a `t="inlineStr"` cell is parsed by `parseRichString`, so inline rich text yields runs.
- `ReadOnlyCell.value` documents that rich text is among the values it can hold, and points at `cellValueAsString` for callers that want only the text.
- `loadWorkbookStream` documents the one difference that remains.

## What this deliberately does not change

The streaming reader reports a cell as empty for a `t="s"` index with no shared string behind it and for a `t="b"` value outside `xsd:boolean`, where `loadWorkbook` throws. That looked like part of the same divergence, but `tests/worksheet/boolean-cell-lexical-forms.test.ts` pins it (`loadWorkbookStream stays lenient on the same input`), so it is a decision rather than an oversight, and a defensible one: a pass over a sheet too large to model should not lose every row after one corrupt cell. Left as it is and documented on `loadWorkbookStream`, so the next reader of that code does not take it for an accident. Worth a separate discussion if the maintainers want the two aligned.

## Testing

```sh
pnpm vitest run tests/streaming/rich-text-agreement.test.ts
```

- Each case reads one set of bytes through both readers and asserts the values are equal, rather than asserting an expected shape per reader. That is the property at issue, and it keeps passing only while the two stay aligned.
- Covers shared strings and inline strings. The inline fixture is built by rewriting the modelled writer's output, since no public API asks for inline strings directly: the write-only path uses them only after 100,000 shared entries.
- One case pins `_xHHHH_` decoding inside an inline run. The old code unescaped each `<t>` as it closed, specifically so a chunk boundary could not split a sequence and two adjacent runs could not form one between them; building the subtree has to keep that, and this is what proves it does.
- One case writes through `createWriteOnlyWorkbook` and reads through `loadWorkbookStream`, which is the shape a streaming export-then-verify pipeline has.
- All 4 fail on `main`.
- Full suite: no change against `main` (the 42 failures on both sides are the conformance tests that need `xmllint`, which is absent on this machine).

## Breaking changes

Pre-1.0 behaviour change, flagged as a minor in the changeset rather than a patch.

A streaming consumer that treated every string cell as a `string` will now see an object for a rich-text cell. `typeof value === 'string'` stops being true for those cells, and `String(value)` would give `[object Object]`. The fix for such a caller is `cellValueAsString` from `@office-kit/xlsx/cell`, which is what `loadWorkbook` consumers already use. Called out in the changeset.

## Checklist


baseballyama (approved):
Thank you so much! Reusing `parseRichString` for the streamed `<is>` instead of keeping a second CT_Rst definition is the right shape, and the CDATA fix in `iterParse` is a good find on its own. I pushed one commit updating `docs/write-only-strings.md`, which still claimed streaming reads flatten rich text.

PR #198, opened 2026-09-22
chore(perf): gate read throughput per machine instead of in cells/s

kibertoad:
## Summary

The read throughput gate asserted a fixed 120,000 cells/s. The CI fleet does not read at a fixed speed: across the last 15 runs the same test reported between 88,806 and 239,223 cells/s on unchanged code, and it failed a PR whose diff was two comments. The gate now divides its throughput by a machine probe and asserts the ratio, so the runner drops out of the number.

## Motivation

Opened without an issue. Found when `perf` failed on #190, whose head commit changed two docblocks and nothing else.

The spread is not "the machine was busy". On the failing run all three read iterations were uniformly slow, 81.0k / 88.8k / 81.4k, so the best-of-3 the test already does could not rescue it. In the same job on the same machine the write gate reported 673,794 cells/s, inside its usual 673k to 694k band, and the runner image was identical to the passing runs (`ubuntu-24.04 / 20260907.300.1`).

Read and write differ in what they stress. The read test materialises 300,000 cells into a model; the write test streams and stays small. Allocation throughput varies far more across the fleet than compute throughput does, which is why one gate swings 2.7x while the other holds inside 3%.

Recent history, from the `[perf]` lines in the job logs:

| runs | read gate | write gate |
| --- | --- | --- |
| 14 passing | 139k to 239k cells/s | 673k to 694k cells/s |
| the failing one | 88.8k | 673.8k |

## Changes

- `tests/perf/machine-probe.ts`: builds the shape the worksheet model builds, a `Map` per row of small objects with a string field, and reports allocations per second. It touches no library code on purpose. If it called into `src/`, a regression in the read path would slow the probe too, the ratio would not move, and the gate would go blind to the thing it exists to catch.
- `read-throughput.test.ts` gates on cells/s per probe unit instead of cells/s. The raw number stays in the log so the historical band remains readable.
- The `perf` job runs the gate twice. The probe covers a runner that is uniformly slower than the median; a stall confined to one measurement window is a different failure, and the second attempt covers that. A real regression fails both.

## Testing

Six local samples, same machine, deliberately perturbed (idle, under six allocation hogs, and warm):

| raw cells/s | probe allocs/s | ratio |
| --- | --- | --- |
| 155,745 to 238,757 (1.53x spread) | 9.19M to 14.38M | 0.0155 to 0.0182 (1.17x spread) |

That is the claim in one line: a 53% swing in the raw metric shows up as 17% in the gated one.

First CI run on this branch: 147,796 cells/s raw, which is the historical median, against a probe of 7,716,231 allocs/s, so 0.0192 per probe unit. Set beside the laptop samples that is a 1.62x spread in the raw number against 1.24x in the gated one, across two different machines.

## Breaking changes

None. Test and CI only, no runtime code, so no changeset.

## Calibration, and what is still unproven

The floor is 0.01 against the 0.0155 to 0.0182 this laptop reports, which is deliberately loose. The ratio's stability across the CI fleet is an argument from construction plus six samples on one machine; the data point I actually want is a ratio measured on a runner like the one that produced 88.8k, and that shape appears about once in fifteen runs. Until one is on record, a tight floor would be a guess dressed as a measurement.

So this lands loose and gets tightened from the CI band once a few runs have reported. The write gate keeps its absolute 500k floor: its spread is under 3% and it has never been the flaky one, so widening the change to it would be churn.

## Checklist

- [x] I have added or updated documentation where user-visible behavior changed. No user-visible behavior changed here; the reasoning lives in the test and probe comments.
- [x] If this is a user-visible change, I have run `pnpm changeset` and committed the result. Not user-visible: test and CI only, so no changeset.
- [x] If this is a breaking change, I have flagged it above and the changeset is marked accordingly. Not a breaking change.


baseballyama (approved):
Thank you so much! Great diagnosis — isolating that only the read path is allocation-heavy explains why the write gate stayed stable. Keeping the probe out of `src/` is the right call. The 0.01 floor is loose for now, but that's documented; we can tighten it once the CI band settles.

